Issue #10, opened 2021-06-01
Improve install process


bbkane:
Hello,

I'm trying to install `templ`, and I have a couple of suggestions to make it easier to install:

# Include an "Install" section in README.md

I might have missed a section, but I didn't see how to install `templ` from the README, only how to build locally. It can be a really short section, but also really helpful :) If you'd like inspiration, I'm pretty happy with how [my Install section](https://github.com/bbkane/kvcrutch/blob/master/README.md#install) turned out.

# Make `templ` easily `go install`able

I was able to install `templ` with `go install github.com/a-h/templ/cmd`, but because the last directory is named `cmd`, the binary is now named `cmd` on my system. If you moved it to `github.com/a-h/templ/cmd/templ`, I would be able to run `go install github.com/a-h/templ/cmd/templ` and end up with a binary named `templ`

# Use GoReleaser's HomeBrew support

GoReleaser can support [generating Homebrew formulas](https://goreleaser.com/customization/homebrew/) so MacOS users can do something like `brew install a-h/tap/templ` if they wish. They won't even need `GOPATH/bin` in their `PATH`, because Homebrew will manage paths. I use this functionality for a little CLI I wrote and it works well once you get the tap repo and access token build- very much set up and forget. Feel free to take inspiration from my [`.goreleaser.yml`](https://github.com/bbkane/kvcrutch/blob/master/.goreleaser.yml)

bbkane:
I'm happy to make PRs to effect these changes should you decide you'd like some/all of them implemented.

a-h:
Brilliant, thanks a lot for thinking about this.

For some reason, I didn't think of the `go install` workflow at all for installation. I really like your idea of moving the command so the executable name is correct when you run `go install`.

Since `templ` is a tool for Go programmers, users will definitely have Go installed, I think that would make homebrew unnecessary, what do you think?

Reading through your docs, I noticed that you've written that using `go install` means that you lose version information. I was a bit worried about this, because I thought it would be hard to chase down bugs if people couldn't say what version of templ they were using, so I did a bit of research.

It turns out that you can get the Go module version at runtime using https://pkg.go.dev/runtime/debug#ReadBuildInfo and also from the Go command line tools using the `go version -m <binary>` command (https://golang.org/ref/mod#go-version-m). This might not the case in all Go version, but it's definitely the case in Go 1.16.

I've updated templ to get the Go module version instead of using `ldFlags` passed in by the goreleaser build in this commit: 

[...]

PR #13, opened 2021-06-06
Let goreleaser manage versioning via ldflags

steinfletcher:
I noticed running `templ version` prints `devel` even when using a release downloaded from github releases.

Since goreleaser sets main.version via ldflags this should ensure it's picked up. I think it's possible to use ReadBuildInfo with goreleaser but requires some additional config.

I tested this using the dry run feature of goreleaser which seemed to work.

```
make build-snapshot
./dist/templ_darwin_amd64/templ version
v0.0.98-next
```

### DIFF
--- a/cmd/templ/main.go
+++ b/cmd/templ/main.go
@@ -4,20 +4,13 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"runtime/debug"
 
 	"github.com/a-h/templ/cmd/templ/fmtcmd"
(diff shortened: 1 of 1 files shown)


a-h:
Thanks for spotting that. I didn't spot that binary builds weren't getting versioned properly any more.

The background is here - https://github.com/a-h/templ/issues/10

I think I need to update the README to make it clear that you can now install a binary build from the Github releases, _or_ install from source with`go install github.com/a-h/templ/cmd/templ@latest`

PR #32, opened 2021-11-08
fix: enable DocumentFormattingProvider

mattn:
it is disabled in default.


a-h:
Another improvement, thanks!

PR #33, opened 2021-11-08
fix: use dir

mattn:
probably dir should be used.


a-h:
Yes, that looks right. I'll give it a test! Thanks!

PR #39, opened 2022-03-17
Write files to buffer rather than directly to files

joerdav:
After some profiling of the generate command I had some interesting results.

I created a 10MB templ file which is the largest allowed then profiled the generation of it.

Ignoring any small effects profiling had on the speed it took around 2 minutes 50 seconds consistently.
Peaking at around 20MB allocated memory.

According to pprof the majority of this was time spent in `syscall`.
Using a file as an io.Reader was calling syscall a lot, which is slow.

So I tried writing to a buffer then writing the buffer and reduced this drastically.

It took about 8 seconds with a similar memory peak.

### DIFF
--- a/cmd/templ/generatecmd/main.go
+++ b/cmd/templ/generatecmd/main.go
@@ -1,6 +1,7 @@
 package generatecmd
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


joerdav:
My system info:

Software:
    System Software Overview:
      System Version: macOS 12.3 (21E230)
      Kernel Version: Darwin 21.4.0

Hardware:
    Hardware Overview:
      Model Name: MacBook Pro
      Model Identifier: MacBookPro16,1
      Processor Name: 6-Core Intel Core i7
      Processor Speed: 2.6 GHz
      Number of Processors: 1
      Total Number of Cores: 6
      L2 Cache (per Core): 256 KB
      L3 Cache: 12 MB
      Hyper-Threading Technology: Enabled
      Memory: 32 GB

a-h:
Looks great, thanks.

PR #51, opened 2022-06-08
feat: add html-style call templates and block templates

joerdav:


joerdav:
@a-h I'm happy that this works as expected now.
Would be good to get your eyes over it first and make sure it's acceptable.
I left `callTemplatesExpressions` in for backwards compatability, was this right?

parser/v2/templelementparser.go
```diff
@@ -0,0 +1,134 @@
...
+	// Read the required closing brace.
+	if ie := closeBraceWithOptionalPadding(pi); !ie.Success {
+		return parse.Failure("templElementParser", newParseError(fmt.Sprintf("@%s: missing end (expected '}')", r.Expression.Value), from, NewPositionFromInput(pi)))
+	}
+	return parse.Success("templElementParser", r, nil)
+}
+
+var templSelfClosingElementExpression templSelfClosingElementExpressionParser
+
+type templSelfClosingElementExpressionParser struct{}
```
a-h, line 73: With this one, is there a need for it now?

I assume that `{! template() }` still works for templates that don't have children.

Is it that `@template()` and `@template() {\n}` can also be used?

parser/v2/templelementparser_test.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+func TestTemplElementExpressionParser(t *testing.T) {
+	var tests = []struct {
+		name     string
+		input    string
+		expected TemplElementExpression
+	}{
+		{
+			name:  "tempelement: simple",
```
a-h, line 17:
```suggestion
			name:  "templelement: simple",
```

parser/v2/templelementparser_test.go
```diff
@@ -0,0 +1,148 @@
...
+							Index: 14,
+							Line:  0,
+							Col:   14,
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "tempelement: simple, block with text",
```
a-h, line 38:
```suggestion
			name: "templelement: simple, block with text",
```

parser/v2/templelementparser.go
a-h: Ah, yes, I can see that in the tests.

parser/v2/templelementparser_test.go
```diff
@@ -0,0 +1,148 @@
...
+				},
+				Children: []Node{
+					Whitespace{Value: "\t"},
+					Text{Value: "some words"},
+					Whitespace{Value: "\n"},
+				},
+			},
+		},
+		{
+			name: "tempelement: simple, block with anchor",
```
a-h, line 66:
```suggestion
			name: "templelement: simple, block with anchor",
```

parser/v2/templelementparser_test.go
```diff
@@ -0,0 +1,148 @@
...
+					Whitespace{Value: "\t\t\t"},
+					Element{Name: "a", Attributes: []Attribute{
+						ConstantAttribute{"href", "someurl"},
+					}},
+					Whitespace{Value: "\n\t\t"},
+				},
+			},
+		},
+		{
+			name: "tempelement: simple, block with templelement as child",
```
a-h, line 96:
```suggestion
			name: "templelement: simple, block with templelement as child",
```

runtime.go
```diff
@@ -33,6 +33,24 @@ func (cf ComponentFunc) Render(ctx context.Context, w io.Writer) error {
...
+	return context.WithValue(ctx, contextKeyChildren, nil)
+}
+
+func GetChildren(ctx context.Context) Component {
+	if component, ok := ctx.Value(contextKeyChildren).(*Component); ok && component != nil {
+		return *component
+	}
+	return ComponentFunc(func(ctx context.Context, w io.Writer) error { return nil })
+}
+
```
a-h, line 53: Maybe it's a good idea to only create the NopComponent once. Also swapped the logic to have happy path down the left.

```suggestion
// NopComponent is a component that doesn't render anything.
var NopComponent = ComponentFunc(func(ctx context.Context, w io.Writer) error { return nil })

// GetChildren from the context.
func GetChildren(ctx context.Context) Component {
	component, ok := ctx.Value(contextKeyChildren).(*Component)
	if !ok || component == nil {
		return NopComponent
	}
	return *component
}

```

a-h (changes requested):
Thanks, this is looking good. I've suggested a couple of minor bits and pieces. Nothing major.

turbo/stream_templ.go
```diff
@@ -8,10 +8,12 @@ import "github.com/a-h/templ"
 import "context"
 import "io"
 
-func actionTemplate(action string, target string, template templ.Component) templ.Component {
+func actionTemplate(action string, target string) templ.Component {
 	return templ.ComponentFunc(func(ctx context.Context, w io.Writer) (err error) {
 		ctx, _ = templ.RenderedCSSClassesFromContext(ctx)
 		ctx, _ = templ.RenderedScriptsFromContext(ctx)
+		var_1 := ctx
```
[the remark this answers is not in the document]
joerdav: This was to prevent children being used in components that are nested, I separated it into its own context so that only the { children... } further down can access that context.

If that makes sense.

Otherwise if another templElement was rendered in this component it would get the children passed in, which is bad.

joerdav:
Was thinking of updating the "Write" function of the calltemplate expression to write out the new syntax. So templ fmt would migrate.

parser/v2/templelementparser.go
```diff
@@ -0,0 +1,134 @@
...
+func (p templBlockElementExpressionParser) Parse(pi parse.Input) parse.Result {
+	var r TemplElementExpression
+
+	// Check the prefix first.
+	prefixResult := parse.Rune('@')(pi)
+	if !prefixResult.Success {
+		return prefixResult
+	}
+
+	// Once we have a prefix, we must have an expression that returns a template.
```
a-h, line 28: Thinking about this, this is a bit harsh as a parser requirement.

If you have:

```html
<div>
@twitterhandle
</div>
```

This parser will throw an error. I think it would be better if it looked for `@dhjhjkf {` as the start. I'll change it though.

parser/v2/templelementparser.go
```diff
@@ -0,0 +1,134 @@
...
+	// Once we've had the start of a for block, we must conclude the block.
+
+	// Eat newline.
+	if lb := newLine(pi); lb.Error != nil {
+		return lb
+	}
+
+	// Node contents.
+	from = NewPositionFromInput(pi)
+	pr = newTemplateNodeParser(parse.Or(elseExpression.Parse, closeBraceWithOptionalPadding)).Parse(pi)
```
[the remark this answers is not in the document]
a-h: Are you sure about this? It doesn't seem right to have the `elseExpression` in here. Must be a copy/paste error?

PR #55, opened 2022-06-20
fix: don't self-close empty script tags

brentd:
Hi, I encountered a bug in v2 of templ where `<script src="..."></script>` elements are being self-closed by the formatter and end up looking like this:

```html
<script src="..."/></script>
```

I added a reproducing test case and fixed it, but I'm not sure if my implementation change makes sense.

Thanks for the review and thanks for making templ, I love the concept!


a-h (approved):

a-h:
Thanks for the contribution! 👍🏻

PR #62, opened 2022-09-28
feat: pool byte buffers during template rendering

joerdav:
I've taken a look at pooling byte buffers to reduce allocations.

I've put the `sync.Pool` behind 2 functions which doesn't seem to have a significant impact on performance.

This looks to have given a 30% speed improvement during generation on my machine.
I believe this performance improvement will only be realised when rendering many templates, not when rendering a single large template.

#### Old
```
goos: darwin
goarch: amd64
pkg: github.com/a-h/templ/benchmarks/templ
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkCurrent-12              1197956               980.8 ns/op          1576 B/op         12 allocs/op
BenchmarkIOWriteString-12       11108792               104.7 ns/op           352 B/op          2 allocs/op
PASS
ok      github.com/a-h/templ/benchmarks/templ   3.395s
```

#### New
```
goos: darwin
goarch: amd64
pkg: github.com/a-h/templ/benchmarks/templ
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkCurrent-12              1696507               692.2 ns/op           568 B/op          7 allocs/op
BenchmarkIOWriteString-12       11542252               103.3 ns/op           352 B/op          2 allocs/op
PASS
ok      github.com/a-h/templ/benchmarks/templ   3.348s
```


generator/generator.go
```diff
@@ -336,7 +336,11 @@ func (g *generator) writeTemplate(t parser.HTMLTemplate) error {
 		}
 		{
 			indentLevel++
-			// _, err = io.Copy(w, buf)
+			// templ.ReleaseBuffer(templBuffer)
```
a-h, line 339: Shouldn't the buffer be released _after_ copying the results?

joerdav: Yes, definitely.
Updated that.

PR #97, opened 2023-06-06
just idea - watch ;)

blinkinglight:


blinkinglight:
any thoughts on this? :)

a-h:
Thanks for your work on this. I tried out the `fsnotify` and had no success with it - it wouldn't produce any events at all on my machine.

I also tried out rjeczalik/notify and had the same issue - see this branch - https://github.com/a-h/templ/tree/hot_reload_watcher

I've implemented a fairly simple `os.WalkDir` implementation with backoff to control the CPU usage at #130

It turns out that filesystem notifications are a bit of a mess, still!

blinkinglight:
p.s. in golang, you can speficy build tags ( so lets say fsnotify_osx.go compiles only in osx build )

PR #124, opened 2023-08-16
fix: replace new lines in constant attributes with escaped new lines

### LINKED ISSUE #115: Multi-line attributes are not supported
Brought to my attention by @c-nv-s in the HTMX discord.

When writing hyperscript it is common to have an attribute with multiple lines, such as:

```
data-script="on click
                do something
             end"
```

In this case an error is returned "string literal not terminated"

joerdav:
Fixes #115


joerdav:
The attributes were being parsed properly, however when generating into _templ.go files the strings were being broken.

PR #130, opened 2023-08-24
feat: hot reload, fixes #82

### LINKED ISSUE #82: proposal: build in hot-reload (templ generate --watch)
The documentation suggests using [air](https://github.com/cosmtrek/air) - https://templ.guide/commands-and-tools/hot-reload

But it's installing another tool, and having to configure it. Running `templ generate --watch` would be a better user experience.

It would make #80 irrelevant, since it would be able to maintain internal state such as the last-updated time. https://github.com/fsnotify/fsnotify seems to be the library to use.

Since templ doesn't know how the web server is configured, it can't easily inject client side code to force a refresh.

But... templ could ship with HTTP middleware that implements hot reload. The middleware would be able to set the target directory that contains the templates and use fsnotify to watch it, or figure it the target directory from `os.Args`, run `templ generate` and then start a replacement of itself based on the `os.Args`. Since it's HTTP middleware, it could parse the output HTML and inject stuff into the page.

Alternatively, it could break the problem into two parts... so that `templ generate --watch` starts a web server on localhost which controls hot reload, and it also ships a middleware component that injects a script to connect to the `templ generate --watch` port (e.g. http://localhost:14734/) and deals with CORS.

a-h:


joerdav:
Worth a look, github have a fork of the browser package imported in this PR that is actively maintained and updated:
https://github.com/cli/browser

a-h:
I'm getting fairly happy with this now. I've been doing some work to reduce the CPU utilisation, since I noticed it was using a lot of CPU while watching.

I added profiling, and noted that walking the directory every 250ms caused a lot of syscalls, and switched to `WalkDir`.

However, I think it's still too much. Next, I thought that I'd try out a phased approach:

* On start, walk all directories and compile a list of files.
* Every 250ms, check the existing set of found files for changes, and maybe do a backoff so that it will wait for up to 1000ms (not sure on the right value yet) to check files that haven't changed recently.
* Every 1000ms-2000ms, re-walk all directories and update the list of found files.

joerdav:
It would be more work to maintain, but you could use a separate technique for systems that support inotify rather than polling.

Or if you're willing to take on another dep:
https://github.com/fsnotify/fsnotify

a-h:
I tried out using fsnotify and friends, but couldn't get it to work well on my machine.

Here's an abandoned branch - https://github.com/a-h/templ/tree/hot_reload_watcher

I'm not sure I want to support the weird errors I'll get with a filesystem watcher implementation, since it requires me to have a machine for each OS and to test it out.

PR #138, opened 2023-09-04
fix: solve auto-import for completions that require them

agalbenus:


cmd/templ/lspcmd/proxy/server.go
```diff
@@ -307,6 +307,8 @@ func (p *Server) ColorPresentation(ctx context.Context, params *lsp.ColorPresent
 	return
 }
 
+var pgkFromImportDetail = regexp.MustCompile(`"([^"]+)"`)
```
a-h, line 310:
```suggestion
var pkgFromImportDetail = regexp.MustCompile(`"([^"]+)"`)
```

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -344,21 +346,36 @@ func (p *Server) Completion(ctx context.Context, params *lsp.CompletionParams) (
...
 			}
-			imp := addImport(doc.Lines, item.Detail)
-			te := lsp.TextEdit{
-				Range: lsp.Range{
-					Start: lsp.Position{Line: uint32(imp.LineIndex), Character: 0},
-					End:   lsp.Position{Line: uint32(imp.LineIndex), Character: 0},
-				},
-				NewText: imp.Text,
+
+			te := handleImportTextEdit(doc, &item, item.Detail)
```
a-h, line 350:
```suggestion
			te := handleImportTextEdit(doc, item.Detail)
```

The item isn't actually used by `handleImportTextEdit`.

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -344,21 +346,36 @@ func (p *Server) Completion(ctx context.Context, params *lsp.CompletionParams) (
...
+			item.AdditionalTextEdits = []lsp.TextEdit{te}
+		}
+		if strings.Contains(item.Detail, "(from \"") {
+			doc, ok := p.TemplSource.Get(string(templURI))
+			if !ok {
+				continue
 			}
+			pkg := pgkFromImportDetail.FindStringSubmatch(item.Detail)[0]
```
a-h, line 358: Is there a possibility that there's no match, and this panics?

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -344,21 +346,36 @@ func (p *Server) Completion(ctx context.Context, params *lsp.CompletionParams) (
...
+			te := handleImportTextEdit(doc, &item, pkg)
```
a-h, line 359:
```suggestion
			te := handleImportTextEdit(doc, pkg)
```

item isn't used by handleImportTextEdit.

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -344,21 +346,36 @@ func (p *Server) Completion(ctx context.Context, params *lsp.CompletionParams) (
...
 			item.AdditionalTextEdits = []lsp.TextEdit{te}
 		}
 		result.Items[i] = item
 	}
 	return
 }
 
+func handleImportTextEdit(doc *Document, item *lsp.CompletionItem, pkg string) lsp.TextEdit {
```
a-h, line 367:
```suggestion
func handleImportTextEdit(doc *Document, pkg string) lsp.TextEdit {
```

Remove unused param.

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -344,21 +346,36 @@ func (p *Server) Completion(ctx context.Context, params *lsp.CompletionParams) (
...
-				Range: lsp.Range{
-					Start: lsp.Position{Line: uint32(imp.LineIndex), Character: 0},
-					End:   lsp.Position{Line: uint32(imp.LineIndex), Character: 0},
-				},
-				NewText: imp.Text,
+
+			te := handleImportTextEdit(doc, &item, item.Detail)
+			item.AdditionalTextEdits = []lsp.TextEdit{te}
+		}
+		if strings.Contains(item.Detail, "(from \"") {
```
a-h, line 353: Is there anything else we could do here around checking the completion kind? Just to limit the likelihood that this regexp matches with something else and triggers when it shouldn't?

a-h (changes requested):
This is looking great. Thanks for jumping in and finishing it off!

cmd/templ/lspcmd/proxy/server.go
agalbenus: what we can do here is to replace the simple `strings.Contains` with another, more explicit, regex expression. Something like `^(type|func|const|var)?\s?\(from\s".+"\)$`. This will also avoid panics on `pkg := pkgFromImportDetail.FindStringSubmatch(item.Detail)[0]` because they complement each other

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -346,18 +347,19 @@ func (p *Server) Completion(ctx context.Context, params *lsp.CompletionParams) (
...
-
 			te := handleImportTextEdit(doc, item.Detail)
 			item.AdditionalTextEdits = []lsp.TextEdit{te}
-		}
-		if strings.Contains(item.Detail, "(from \"") {
-			doc, ok := p.TemplSource.Get(string(templURI))
-			if !ok {
-				continue
+		} else {
+			m := completionWithImport.FindStringSubmatch(item.Detail)
```
agalbenus, line 353: it seems that on the go side, any kind besides module can have the package embedded in the detail field 

[also it seems that only completion](https://github.com/golang/tools/blob/21090a2aa8d3719d5a5d4264e41696529522f0bd/gopls/internal/lsp/source/completion/format.go#L179)

I think this should do it. Let me know what do you think

cmd/templ/lspcmd/proxy/server.go
agalbenus: check last commit

cmd/templ/lspcmd/proxy/server.go
agalbenus: I think that last commit excludes any panic scenarios

a-h (approved):

PR #143, opened 2023-09-07
fix: don't error if no open brace is found on control flow expressions

### LINKED ISSUE #141: Bug: <div>: expected end tag not present or invalid tag contents
Take the following code:

			<div class="row">
				<div> Price </div>
				<div> Free </div>
				<div> for X month </div>
				<div> for X month </div>
			</div>

Looks perfectly valid, right? Yet, we get the "expected end tag not present or invalid tag contents" error.

No hidden characters, ... every div is properly closed. Guess where it fails on ;)

If your guessed "for", you win a price! For some reason "for" inside the div, messes with the regex of templ, and it creates an unrelated error. Remove the two times "for" and code compiles perfectly with no errors.

joerdav:
fixes #141

### DIFF
--- a/docs/docs/05-server-side-rendering/01-creating-an-http-server-with-templ.md
+++ b/docs/docs/05-server-side-rendering/01-creating-an-http-server-with-templ.md
@@ -76,7 +76,7 @@ The output will always be the date and time that the web server was started up, 
 2023-04-26 08:40:03.421358 +0100 BST m=+0.000779501
 ```
 
-To display the current time, we could update the component to use the `time.Now()` function itself, but this would limit the reusability of the component. It's better when components are take parameters for their display values.
+To display the current time, we could update the component to use the `time.Now()` function itself, but this would limit the reusability of the component. It's better when components take parameters for their display values.
 
 :::tip
 Good templ components are idempotent, pure functions - they don't rely on data that's not passed in as parameters. As long as the parameters are the same, they always return the same HTML - they don't rely on any network calls or disk access.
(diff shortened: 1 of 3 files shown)


joerdav:
@a-h Is this how you would expect this to have been fixed?

a-h:
Very close. I was thinking about attempting to search for a greater number of symbols before templ complains about a malformed `for` or `if`

For example. The `for` statement must be constructed in a number of ways.

```go
package main

import "fmt"

func main() {
	var x int
	for ; x < 10; x++ {
		print(x)
	}
	fmt.Println()
	for y := 0; y < 10; y++ {
		print(y)
	}
	for {
		break
	}
	fmt.Println()
	var y int
	for y = 0; y < 10; y++ {
		print(y)
	}
}
```

https://go.dev/play/p/RjaP5Hj5HkV

I was thinking of still complaining about the `for` statement not being closed if there's a semicolon in the line. However, that would simply reduce the times that the spurious error occurs.

My concern was developers accidentally turning a malformed `for` loop into printing `for x := 0; x < checkSomething(); x ++` to the HTML.

So I was thinking about ways to reduce the number of spurious errors while providing some protection against forgetting to close a for loop start expression.

joerdav:
Makes sense. It's finding that line between errors that stop someone from writing certain things in an element and allowing broken for loops to be written.

I guess where it gets harder is `if` and `switch` statements, they are unlikely to contain any discerning character between the keyword and the open-brace.

PR #144, opened 2023-09-10
Return error when generation fails

veggiemonk:
This caught me by surprise when using `air` to auto-rebuild.

The `file.templ` was invalid and `templ generate` does not fail, it just prints the error. It caused the logical AND to succeed and rebuild while I kept working on the project and wondering why I'm not seeing my changes.

```toml
cmd = "templ generate && go build -o ./tmp/main ."
```

With this changes, it fails and the server dies, which is what I was expecting. 
Maybe it is just me, let me know what you think.

Cheers

### DIFF
--- a/cmd/templ/generatecmd/main.go
+++ b/cmd/templ/generatecmd/main.go
@@ -121,7 +121,7 @@ func runCmd(ctx context.Context, args Arguments) (err error) {
 			if errors.Is(errs[0], context.Canceled) {
 				return errs[0]
 			}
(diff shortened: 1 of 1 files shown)


a-h:
Thanks, sorry for that. I missed that it used to return a non-zero exit code when template generation failed, but I broke it when I implemented a basic hot reload (it doesn't use a filesystem watching library because non of them worked well across all platforms).

I would be interested in your thoughts on the built-in hot reload, since you're using hot reload as part of your workflow: https://templ.guide/commands-and-tools/hot-reload

PR #147, opened 2023-09-14
chore: Use static distroless image

KasonBraley:
Since CGO is not required for this example, we can use the distroless static-debian11 image. This reduces the size of the image by 18 MB. Although this is just an example, I figured it would be nice show the usage of this static image in the docs.

```
counter-basic-static latest  635b8094b79a   10 seconds ago   10.6MB
counter-basic        latest  047d3cd36484   2 minutes ago    28.6MB
```


a-h:
Yes, you're right. I'm all for having smaller Docker images! Thanks for your contribution. 👍🏻

PR #148, opened 2023-09-14
docs: Add godoc badge to README

KasonBraley:
I think it would be useful to have a godoc badge in the README for easy access to the package documentation. If approved, please move it to where ever you like. I just put it towards the top like most README's, although it could make sense to put it under the documentation part.

Badge created via https://pkg.go.dev/badge/.


a-h:
Thanks for the contribution!

Issue #150, opened 2023-09-15
CSS expressions middleware wants type templ.ComponentCSSClass & not templ.CSSClass


topi314:
I tried to use https://templ.guide/syntax-and-usage/css-style-management/#css-middleware to combine all my css expressions into a single stylesheet, but templ generates css expressions with the return type of `templ.CSSClass`
example:

```go
css ErrorDetails() {
	text-align: start;
}
```
->
```go
func ErrorDetails() templ.CSSClass {
	var templCSSBuilder strings.Builder
	templCSSBuilder.WriteString(`text-align:start;`)
	templCSSID := templ.CSSID(`ErrorDetails`, templCSSBuilder.String())
	return templ.ComponentCSSClass{
		ID:    templCSSID,
		Class: templ.SafeCSS(`.` + templCSSID + `{` + templCSSBuilder.String() + `}`),
	}
}
```

now I can't pass these classes into `templ.NewCSSHandler(ErrorDetails())` since it wants `...templ.ComponentCSSClass`

from what I can tell this requires a simple fix in here:
https://github.com/a-h/templ/blame/b6358024ebd0c83257331b6f08053d4a946a2b76/generator/generator.go#L178

a-h:
Hi @topi314 - thanks a lot for looking into this, and sorry for the inconvenience there.

It seems there was a missing spot in the testing. The unit test covered the behaviour, but there was no integration of real usage.

I tried your PR, but it broke the snapshot tests since it resulted in CSS not being rendered. Rather than change the generated type that's output, I tried to stick with `CSSClass` and use a type check in the middleware code instead. Since `CSSClass` is more general (interface type), it shouldn't break any code that's used the variadic method of adding values, but it would break code where people have used an array of `ComponentCSSClass` items since Go can't convert between the two.

I've also added a unit test that describes the expected behaviour so there isn't a future regression. 

https://github.com/a-h/templ/pull/152

I'll sleep on this change before comitting, since it affects the public API of templ. 😁 

Does that work for you?

topi314:
sounds good to me!
thanks for the quick response

topi314:
I would love for CSS support to be expanded beyond classes, it's annoying to manually bundle or import a dozen of stylesheets. Pehaps in component `<style>` tags could also be added to the middleware?

a-h:
[...]

PR #151, opened 2023-09-15
fix wrong css expression return type 

topi314:
this should resolve https://github.com/a-h/templ/issues/150


topi314:
closed in favor of https://github.com/a-h/templ/pull/152

PR #155, opened 2023-09-17
Update 08-template-composition.md

st3fan:
I assume the intent of the first example was to show Left, Middle, Right. This small change fixes that.


a-h (approved):

a-h:
Ah, thanks for that! Much appreciated. 😁 

Thanks for trying out templ, and for taking the time to contribute!

PR #156, opened 2023-09-17
fix: maintain whitespace around templated values

noaoh:
I noticed while following the docs (specifically https://templ.guide/syntax-and-usage/expressions) that for the following template:
```
package main

templ greet(prefix string, p Person) {
  <div>{ prefix } { p.Name }{ exclamation }</div>
}
```
The whitespace between `{prefix}` and `{p.Name}` is not preserved.  This is an attempt to fix that.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -412,6 +412,14 @@ func stripNonCriticalElementWhitespace(input []parser.Node) (output []parser.Nod
 			output = append(output, curr)
 			continue
 		}
+		
+		_, prevIsStringExpr := prev.(parser.StringExpression)
+		_, nextIsStringExpr := next.(parser.StringExpression)
+		if prevIsStringExpr || nextIsStringExpr {
+			// Allow whitespace that comes before or after a template expression.
+			output = append(output, curr)
+			continue
+		}
 	}
 	return
 }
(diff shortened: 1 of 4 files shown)


a-h:
You're absolutely right. Good spot, thanks for that! I checked and saw that people have been putting in `&nbsp;` at work.

This is much better!

a-h:
I took a look at this yesterday, it does look great, but I realised that the `templ fmt` removes whitespace between templated values, so I wanted to think about this change and if/how it interacts with `templ fmt`, hence the pause.

There's an open issue around the formatting behaviour which I need to revisit. I've had a few attempts at it, but haven't managed to get something I'm 100% happy with yet.

From discussion with @joerdav, I think the way we might go with formatting is simply to add additional indentation where it's missing, but do nothing more than that - i.e. no whitespace removal, and then on generation, normalise whitespace according to HTML rules (i.e. collapse runs of whitespace down to a single space).

noaoh:
My thought would be that formatting should remove extra whitespace where it's found, but leave a single character of whitespace. This is so there's no differences between the formatted template, and what gets generated.  Just my two cents.

joerdav:
@noaoh the difficulty being that it seems everyone wants to format their HTML differently in templ, some want all their elements to be displayed on new lines similar to prettier, others want some elements to be inlined and not others.

So I think the best medium is that templ fmt doesn't destroy any whitespace, but maybe can add some in cases where the intent is obvious, like turning the following into a block element:

```html
<!-- input -->
<div>hello
</div>
<!-- becomes -->
<div>
    hello
</div>
```

a-h:
Great job, thanks for sorting that out - it's much appreciated.

The formatting changes were just fixed, and this PR worked perfectly in conjunction.

PR #166, opened 2023-09-24
docs: fix basic syntax example

levidurfee:
Hi, I believe this example should be Go code, not the templ code.

### DIFF
--- a/docs/docs/03-syntax-and-usage/01-basic-syntax.md
+++ b/docs/docs/03-syntax-and-usage/01-basic-syntax.md
@@ -34,12 +34,15 @@ Outside of templ Components, templ files are ordinary Go code.
 ```templ name="header.templ"
 package main
 
+// Ordinary Go code that we can use in our Component.
 var greeting = "Welcome!"
 
+// templ Component
 templ headerTemplate(name string) {
   <header>
     <h1>{ name }</h1>
-    <h2>{ greeting }</h2>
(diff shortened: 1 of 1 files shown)


joerdav:
Actually, I think the original is correct. This section is meant to show that even in a templ file you can write pure go code outside of the `templ` block.

Such as `var greeting = "Welcome!"`

a-h:
Yes, @joerdav is right, it's supposed to show that. But apparently, it's not doing a very good job of explaining, so there's maybe some improvement to the docs to do! 😁 

@levidurfee - what do you think would have made that point clearer? Maybe highlighting the `var greeting = "Welcome!"` and `<h2>{ greeting }</h2>` lines?

levidurfee:
Oh okay, I see what you mean. 

I think something like the following might help make it more clear.

> Outside of templ Components, templ files can contain ordinary Go code.

```
package main

// Ordinary Go code that we can use in our Component.
var greeting = "Welcome!"

// templ Component
templ headerTemplate(name string) {
  <header>
    <h1>{ name }</h1>
    <h2>"{ greeting }" comes from ordinary Go code</h2>
  </header>
}
```

a-h:
Thanks - if you want to update this PR to make that change, go for it, otherwise, shall we close this?

levidurfee:
I'll update the PR

levidurfee:
@a-h I've updated the PR

a-h:
Thanks for that!

Issue #87, opened 2023-05-08
templ fmt indents weirdly if @component is put before/after text 


a-h:
The following snippet:

```html
templ x() {
	<li>
		<a href="/">
			Home
			@Icon("home", Inline)
		</a>
	</li>
}
```

gets formatted to become:

```html
templ x() {
	<li><a href="/">
	Home
	@Icon("home", Inline)
</a></li>
}
```

I don't think this is expected behavior.

(Extracted from #86)

a-h:
Please see https://github.com/a-h/templ/commit/7efc71cfcfead03bbf773850db350e2279ec700b - I've updated the formatting. I hadn't taken into account correct whitespace handling behaviour at the start and end of inline elements.

The `@Icon` was always being interpreted as a block element.

Since `<a>` elements are inline, the content is now moved to a single line, with the whitespace correctly collapsed into a single space between the "words".

```
package main

templ x() {
        <li><a href="/">Home @hello("home")</a></li>
}
```

diamondburned:
This is unrelated to the issue, but is there a way to disable collapsing of inline elements into a single line? It would be really inconvenient to have a really long paragraph be squished into one line instead of being line wrapped properly.

For example:

```
templ x() {
	<p>
		Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
		tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
		veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
		commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
		velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
		occaecat cupidatat non proident, sunt in culpa qui officia deserunt
		mollit anim id est laborum.
	</p>
}
```

shouldn't be squished to become

```
templ x() {
	<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
}
```

a-h:
[...]

PR #167, opened 2023-09-25
updated formatting to suggested style in #87

evanlurvey:
so this is just a proposal at this point. Not sure if we like how I implemented this. I don't like that it broke a ton of parser tests but I don't know where else to put the state in order to be able to format it properly later.


parser/v2/types.go
```diff
@@ -415,29 +417,28 @@ func (e Element) Write(w io.Writer, indent int) error {
 	if err := writeIndent(w, indent, "<"+e.Name); err != nil {
 		return err
 	}
-	var previousWasMultiline bool
```
evanlurvey, line 418: we no longer care about multi line because the parser for conditionals requires a newline after the `{` which would populate IndentAttrs property through parsing logic.

a-h: I think that putting the state into the element type makes a lot of sense. It makes the code neater too.

a-h:
Wow, this looks great. Thanks so much for taking the time to do this.

I'll take a proper look tomorrow!

a-h:
Wonderful work, thanks a lot.

Issue #146, opened 2023-09-12
hot-reload feature: experience report


veggiemonk:
Hey @a-h,

I tried the hot-reload you recommended. (see #144)

## Workflow

```sh
# Install templ
go install github.com/a-b/templ/cmd/templ@latest

air -c .air.toml

# In another terminal

templ generate --watch --proxy="http://localhost:8080"

# modified a .templ file
```

### Issues

Here are the issues I've encountered while trying out templ.

#### Issue 1 - Race condition between the proxy and the server

```log
Generated code for 1 templates with 0 errors in 3.164459ms
Opening URL: http://127.0.0.1:8080
Proxying from http://127.0.0.1:7331 to target: http://127.0.0.1:8080
Proxy to target error: http: proxy error: dial tcp 127.0.0.1:8080: connect: connection refused
Proxy to target error: http: proxy error: dial tcp 127.0.0.1:8080: connect: connection refused
```

Workaround: Delay opening the browser

#### Issue 2 - Filtering on header `Content-Type` is a really bad idea.

`text/html` is not the only valid value for `Content-Type` header. For example, `text/html; charset=utf-8`.

Solution: replace `contentType != "text/html"` with `!strings.HasPrefix(contentType, "text/html")`

#### Issue 3 - The sending of event is triggered at the same time that the server is rebuilding which causes to miss the update.

workaround: delay the sending of the event

#### Issue 4 - Mandatory use of `--cmd` flag which is not really documented what is expected/allowed.

```sh
templ generate --watch --proxy="http://localhost:8080" --cmd="go run ."
```

It was truncated to only one argument, i.e. `go run`. 
Once fixed, the previous `go run .` was not killed in time for the new one to start and got a port conflict.

Solution: just set a placeholder i.e. `--cmd="echo"` and move on.

Note: If a nonexistent command is given, on the first pass, it works fine. On the second, it segfaults.

#### Issue 5 - With all the `time.Sleep` added, it takes more than 5s reload. Very slow.

Gave up.

See all the code in the draft PR #145

Here is the project I used to test the hot-reload: https://github.com/veggiemonk/hot-templ

I should also mentioned that I had some issue with using `--proxy="http://localhost:8080` and had to change it to `--proxy="http://127.0.0.1:8080` because the ReverseProxy used the IPv6 localhost. 🤷 I'm not sure I knew what was happening.

By the way, I'm using M1 Pro with latest OS.

a-h:
[...]

PR #181, opened 2023-09-30
fix: --cmd breaking with >1 argument, kill child processes correctly

headblockhead:
Fixes issue 4 in the #146 experience report.
Maybe add tests for the cli to prevent future issues.

### DIFF
--- a/cmd/templ/generatecmd/run/run.go
+++ b/cmd/templ/generatecmd/run/run.go
@@ -14,7 +14,7 @@ import (
 var m = &sync.Mutex{}
 var running = map[string]*exec.Cmd{}
 
-func KillAll() {
+func KillAll() (err error) {
 	m.Lock()
 	defer m.Unlock()
 	for _, cmd := range running {
@@ -22,12 +22,19 @@ func KillAll() {
 			kill := exec.Command("TASKKILL", "/T", "/F", "/PID", strconv.Itoa(cmd.Process.Pid))
 			kill.Stderr = os.Stderr
 			kill.Stdout = os.Stdout
-			kill.Run()
+			err := kill.Run()
+			if err != nil {
+				return err
+			}
 			continue
 		}
-		syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+		err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+		if err != nil {
+			return err
+		}
 	}
 	running = map[string]*exec.Cmd{}
+	return
 }
 
 func Run(ctx context.Context, workingDir, input string) (cmd *exec.Cmd, err error) {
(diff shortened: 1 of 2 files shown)


cmd/templ/generatecmd/run/run.go
```diff
@@ -2,39 +2,72 @@ package run
...
+				return err
+			}
+			continue
+		}
+		err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+		if err != nil {
+			return err
+		}
+	}
+	running = map[string]*exec.Cmd{}
```
a-h, line 36: Go 1.21 brings in the new "clear" function, but I want to stick to supporting older versions of Go until they're unsupported by the Go team.

cmd/templ/generatecmd/run/run.go
```diff
@@ -2,39 +2,72 @@ package run
...
-			return nil, fmt.Errorf("failed to kill existing process: %w", err)
-		}
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err = KillAll()
+			if err != nil {
+				return
+			}
```
a-h, line 52: Here, we're checking for a return error, but then returning from the go func.

And also, there's only one thing in the waitgroup.

I don't see the point of the go routine and waitgroup, since the code would operate the same without them.

cmd/templ/generatecmd/run/run.go
a-h: So this is good!

cmd/templ/generatecmd/run/run.go
```diff
@@ -2,39 +2,72 @@ package run
...
+		}()
+		wg.Wait()
```
a-h, line 54:
```suggestion
```

cmd/templ/generatecmd/run/run.go
```diff
@@ -2,39 +2,72 @@ package run
...
-		}
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err = KillAll()
+			if err != nil {
+				return
+			}
+		}()
```
a-h, line 53:
```suggestion
```

cmd/templ/generatecmd/run/run.go
a-h, line 52:
```suggestion
		if err = KillAll(); err != nil {
			return
		}
```

a-h (changes requested):

cmd/templ/generatecmd/run/run.go
headblockhead: I have removed the waitgroup in the latest commit, it works with no problems.

a-h (approved):

PR #183, opened 2023-09-30
fix: apply a templ prefix to all variables that templ uses within a template to reduce clashes, fixes #169

### LINKED ISSUE #169: Variable names defined in templates can clash with variables in compiled output
This template fragment generates invalid output because the `err` variable name clashes with the generated variable name:

```
if err := validationErrors.Get("name"); e != nil {
  <label>{ err.Error() }</label> 
}
```

Generates:

```
if err := errs.Get("name"); e != nil {
    _, err = templBuffer.WriteString(...)
   if err != nil {
	return err
  }
}
```

Other variable names used like `ctx` and even `var_1` have the same drawbacks.

Quicktemplate generates var names like `qb422016` that avoid these clashes.

a-h:


joerdav:
Looks good to me. Only thought is maybe would be a degree safer if these vars were generated from a hash of the variable name:

`varName("ctx") // templ_ctx_166362`

Could be overkill? But might reduce the likelihood of a clash by another big chunk. And since it's a hash there's no git thrashing.

PR #194, opened 2023-10-02
doc: add missing () in CSS component example

wdscxsj:
- Add missing `()` for `css` components.
- Remove redundant trailing `;`.

According to the implementation of `RenderCSSItems()`, `className` still works without invocation, but the previous doc follows a theme that `()` is always good (unless the component is used in other clever ways). I think it's better to keep consistent here, or make it clear in the doc when a factory function is also OK.


a-h (approved):
Agreed, thanks!

Issue #197, opened 2023-10-02
cmd: Add an option to output to file or stdout


rowan-gud:
## Problem

It doesn't seem like there is an option to output to stdout or to a file location.

## Reasoning for wanting this feature

Trying to use templ with bazel and add a custom rule and want to do something like
```starlark
genrule(
  name = "index_templ",
  srcs = ["index.templ"],
  outs = ["index_templ.go"],
  cmd = "templ generate -f $< -o $@", # or templ generate -f $< > $@
)
```

If there is another way to do this or something I am missing in the docs please let me know!

a-h:
I'm not familiar with Bazel.

From reading the docs, it looks like it tries to do some stuff around caching build outputs, but the Go tooling does that already for Go packages.

Can you explain the use case a bit more? Are you trying to use it to remember to run `templ generate` if the `*.templ` files have changed?

rowan-gud:
Yeah so my use case for Bazel is having a mono repo with multiple build artifacts and Bazel handles the dependency tree + build caching etc.

My reasoning for wanting to put templ generate into Bazel is more of convenience of having a single build command. Right now I'm not checking in my `_templ.go` files but I could see having my CI incrementally build those too.

I will take another look at the PR. Thank you!

rowan-gud:
Yeah it would be able to accomplish that with `templ generate -o /path/to/location/dir`

ReneHollander:
To workaround the missing option to specify the output directory for now, the following works for a genrule in a directory called `frontend`:

```
genrule(
    name = "main_templ",
    srcs = ["main.templ"],
    outs = ["main_templ.go"],
    cmd = "$(location @com_github_a_h_templ//cmd/templ:templ) generate -f $< && cp frontend/main_templ.go $@",  # or templ generate -f $< > $@
    tools = ["@com_github_a_h_templ//cmd/templ"],
)
```

It might be possible to improve on this. The hardcoded `frontend/main_templ.go` value would need to be changed if the input file of `main.templ` is moved/renamed.

moshe5745:
+1

moshe5745:
I think it will be just better DX to have all the generated files in one place so they don't mix with our code.

joerdav:
So it looks as though the way forward here is to support printing generated files out to stdout, this will allow greater flexibility for those with different use cases.

This could work similarly to the `fmt` command? `templ generate < file.templ`

jimafisk:
[...]

PR #198, opened 2023-10-02
feat: Add output filename or dir to generate

rowan-gud:
From my preliminary understanding of the codebase this could be a potential solution to [issue-197](https://github.com/a-h/templ/issues/197) . Please let me know what should be changed or done better!

### DIFF
--- a/cmd/templ/generatecmd/main.go
+++ b/cmd/templ/generatecmd/main.go
@@ -117,7 +117,7 @@ func runCmd(ctx context.Context, args Arguments) (err error) {
 		var stat fs.FileInfo
 		stat, err = os.Stat(args.Output)
 		if err != nil {
-			return fmt.Errorf("output directory does not exist: %s", args.Output)
+			return
 		}
 		if !stat.IsDir() {
 			return fmt.Errorf("expected a directory as output when -f is not provided")
@@ -141,7 +141,7 @@ func runCmd(ctx context.Context, args Arguments) (err error) {
 	var firstRunComplete bool
 	fileNameToLastModTime := make(map[string]time.Time)
 	for !firstRunComplete || args.Watch {
-		changesFound, errs := processChanges(ctx, fileNameToLastModTime, args.Path, args.GenerateSourceMapVisualisations, args.WorkerCount, args.Output)
+		changesFound, errs := processChanges(ctx, fileNameToLastModTime, args.Path, args.Output, args.GenerateSourceMapVisualisations, args.WorkerCount)
 		if len(errs) > 0 {
 			if errors.Is(errs[0], context.Canceled) {
 				return errs[0]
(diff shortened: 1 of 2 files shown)


cmd/templ/generatecmd/main.go
```diff
@@ -203,6 +206,14 @@ func processChanges(ctx context.Context, fileNameToLastModTime map[string]time.T
 			return nil
 		}
 		if strings.HasSuffix(path, ".templ") {
+			pathSlices := regexp.MustCompile(`[\\/]`).Split(path, -1)
```
a-h, line 209: Would https://pkg.go.dev/path/filepath#Split be a better choice here? It splits into path and filename, and is aware of the operating system, i.e. Windows vs Unix filesystem path separators.

Even if it wasn't, the regexp compilation would probably be better off outside of the function so that it only happens once per program run.

cmd/templ/generatecmd/main.go
```diff
@@ -251,17 +262,17 @@ func openURL(url string) error {
 	return browser.OpenURL(url)
 }
 
-func processSingleFile(ctx context.Context, fileName string, generateSourceMapVisualisations bool) error {
+func processSingleFile(ctx context.Context, fileName string, generateSourceMapVisualisations bool, outputFileName string) error {
```
a-h, line 265: I think that if we're going to call one file name `outputFileName`, then the `fileName` variable would be better named as `inputFileName`.

Also, I think the order of the parameters should be shifted.

```suggestion
func processSingleFile(ctx context.Context, inputFileName string, outputFileName string, generateSourceMapVisualisations bool) error {
```

cmd/templ/generatecmd/main.go
```diff
@@ -203,6 +206,14 @@ func processChanges(ctx context.Context, fileNameToLastModTime map[string]time.T
 			return nil
 		}
 		if strings.HasSuffix(path, ".templ") {
+			pathSlices := regexp.MustCompile(`[\\/]`).Split(path, -1)
+			fileName := pathSlices[len(pathSlices)-1]
+			outputFileName := ""
+
+			if outputDir != "" {
```
a-h, line 213: What should happen if the output directory doesn't exist? In the current behaviour, that can't happen, because it's operating on files that already exist in that directory.

I don't think this code accounts for templ files in subdirectories at the moment.

For example, if I've got a project layout of:

```
/shared/layouy/header/index.templ
/pages/default/index.templ
/pages/test.templ
```

I think I'd end up with one `index_templ.go` (i.e. I've lost one) and `test_templ.go` in the output directory.

cmd/templ/main.go
```diff
@@ -75,6 +75,7 @@ examples:
 func generateCmd(args []string) {
 	cmd := flag.NewFlagSet("generate", flag.ExitOnError)
 	fileNameFlag := cmd.String("f", "", "Optionally generates code for a single file, e.g. -f header.templ")
+	outputFlag := cmd.String("o", "", "Optionally set an output for the Go code. If -f is set this should be a filename if not this should be a directory.")
```
a-h, line 78:
```
If -f is set this should be a filename, otherwise, it should be a directory.
```

I don't see this being validated or checked anywhere. It seems like I could run `go generate -f test.templ -o test.templ` and overwrite my templ file, or generate something that doesn't match the expected `_templ.go` pattern.

a-h, line 78:
```
If -f is set this should be a filename, otherwise, it should be a directory.
```

I don't see this being validated or checked anywhere. It seems like I could run `go generate -f test.templ -o test.templ` and overwrite my templ file, or generate something that doesn't match the expected `_templ.go` pattern.

a-h (commented):
Let's continue a bit of discussion in https://github.com/a-h/templ/issues/197!

cmd/templ/generatecmd/main.go
rowan-gud: Yeah I think `filepath.Split` would work great here!

skydiver1980 (approved):

innazh:
@rowan-gud , @a-h , any updates on moving through with this change?

rowan-gud:
Hi @a-h sorry I dropped off the face of the earth haha. It looks like a couple things happened, it looks like #628 addresses the underlying issue that I brought up of not being able to change the output of the generation.

It looks like based on the conversation in #735 we may still want the option to output to a file / directory as well.

I'm going to close this PR since I think it's pretty outdated at this point but I would be happy to work together with @jespert to come up with a solution for adding a `-o` flag!

Issue #50, opened 2022-06-05
proposal: write up how to use with https://github.com/gin-gonic/gin


a-h:

cayter:
Trying to get this working with Gin but am facing challenges. Any update on this? Thanks.

cayter:
Got it working.

renderer/templ.go
```
package renderer

import (
	"context"
	"net/http"

	"github.com/a-h/templ"
	"github.com/gin-gonic/gin/render"
)

type TemplRender struct {
	Code int
	Data templ.Component
}

func (t TemplRender) Render(w http.ResponseWriter) error {
	w.WriteHeader(t.Code)
	if t.Data != nil {
		return t.Data.Render(context.Background(), w)
	}

	return nil
}

func (t TemplRender) WriteContentType(w http.ResponseWriter) {
	w.Header().Set("Content-Type", "text/html; charset=utf-8")
}

func (t *TemplRender) Instance(name string, data interface{}) render.Render {
	if templData, ok := data.(templ.Component); ok {
		return &TemplRender{
			Code: http.StatusOK,
			Data: templData,
		}
	}
	return nil
}
```

main.go
```
package main

import (
  "<module>/renderer/templ"
)

func main() {
  r := gin.Default()
  r.HTMLRender = &renderer.TemplRender{}
  
  r.GET("/", (c *gin.Context) {
    c.HTML(http.StatusOK, "", page.Home())
  })
}

```

a-h:
If this became an example in the examples directory, and made its way into the docs, good Hacktoberfest PR.

PR #203, opened 2023-10-03
feat(docs): add integrations and examples

aranw:
Called the section integrations but I guess this could be named something else maybe?

I added the integrations section as 12 so that I wasn't renumbering everything and then possibly conflicting with other contributions

Closes https://github.com/a-h/templ/issues/50


examples/template/html.templ
```diff
@@ -0,0 +1,11 @@
+package template
```
a-h, line 1: I think this is an accidental commit?

examples/template/html_templ.go
a-h, line 1: Same with this?

a-h:
This looks great, thanks for doing that! Just checking whether you meant to commit one of the files.

aranw:
> This looks great, thanks for doing that! Just checking whether you meant to commit one of the files.

Yeah that was a mistake! I've removed it now

a-h (approved):

a-h:
Thanks a lot!

joerdav:
Odd that the PR checks didn't run on this, it looks like the examples were generated with a non-HEAD version of templ.

Maybe we need a contribution guide that says you should run `go run ./cmd/templ generate` rather than `templ generate` before commiting.

Issue #208, opened 2023-10-05
Templ fmt does not format go code


gungun974:
### Introduction (you can skip that part if you want)
So I like go.

I like this project which for me gives even more reason to like go (because the HTML template module seems horrible to me when used in comparison)

**The thing that makes me love this project even more than its syntax is its LSP server**

In general, HTML is error prone to a lot of things. Including just misspell a variable to inject. 
So need to manage all of this without a good support from my editor like JSX is a deal breaker for me.

But fortunately the LSP server of Templ is here to save me and I'm very happy with it ^^

### Problem

However, I have a problem with the formatting aspect of Templ.

Templ is a kind of superset of Go. You can declare in your file package, variable, function like you do in normal go.
In fact the generator just insert the raw go code in generated files.

This work great, specially with the LSP and reuse of Gopls.

However I can't said it's the same for `templ format`.
This command only format Templ syntax when the parser produce a formated code.

So in example like this :

```go
type Link struct {
  	Name string
	Url  string
}

      var a = false;

func test() {
    	log.Print("hoi")

      if (a) {
        log.Fatal("OH NO !")
      }
}
```

`templ fmt` don't produce what `gofmt` would do, this in this case :

```go
type Link struct {
	Name string
	Url  string
}

var a = false

func test() {
	log.Print("hoi")

	if a {
		log.Fatal("OH NO !")
	}
}
```

This is very annoying cause I love the idea of creating variable and function in the same file of the templeting for tiny function but the fact I need to manually format my file make me stress :eyes:

So as in my new tradition of wanting absolutely everything as quickly and without the slightest effort for the maintainers to exist. ***I created a PR to resolve the issue which I'm happy to share*** #209.

a-h:
Thanks for raising this issue, and for providing a PR to fix it. Much appreciated!

PR #209, opened 2023-10-05
feat: make templ fmt, format writen go code in templates

gungun974:
This PR solves the problem as written in the Issue #208 that `go templ` formats Go code which can be written alongside the `templ` keyword.

I'm still new to the go ecosystem but saw what was done in PR #102 with the `go/format` package. 
I think my little modification to format the code with this package when generating the clean string version is good.

Of course I'm open to any improvement.

### DIFF
--- a/parser/v2/types.go
+++ b/parser/v2/types.go
@@ -2,6 +2,7 @@ package parser
 
 import (
 	"fmt"
+	"go/format"
 	"html"
 	"io"
 	"strings"
@@ -139,7 +140,13 @@ type GoExpression struct {
 
 func (exp GoExpression) IsTemplateFileNode() bool { return true }
 func (exp GoExpression) Write(w io.Writer, indent int) error {
-	return writeIndent(w, indent, exp.Expression.Value)
+	data, err := format.Source([]byte(exp.Expression.Value))
+	if err != nil {
+		return writeIndent(w, indent, exp.Expression.Value)
+	}
+
+	_, err = w.Write(data)
+	return err
 }
 
 func writeIndent(w io.Writer, level int, s string) (err error) {
--- a/parser/v2/types_test.go
+++ b/parser/v2/types_test.go
@@ -8,7 +8,7 @@ import (
 )
 
 func TestFormatting(t *testing.T) {
-	var tests = []struct {
+	tests := []struct {
 		name     string
 		input    string
 		expected string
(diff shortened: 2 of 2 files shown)


a-h:
Hey, this looks great. I noticed this the other day, but hadn't had a chance to look into it.

I would like to see a unit test of the behaviour so that we don't accidentally remove this feature and break it in the future.

It shouldn't be much effort to that, just need to write some badly formatted Go code and check that it gets properly indented, some invalid Go code to show that it is left as-is, and some already well formatted Go code to show that it's not adjusted.

jtarchie:
This may result in not needing `writeIndent` execution and tracking. They could just all just be changed to `Write`.

gungun974:
> This may result in not needing `writeIndent` execution and tracking. They could just all just be changed to `Write`.

I just finish to add unit tests.

If I understand what you said, you propose I write this ? :
```go
func (exp GoExpression) Write(w io.Writer, indent int) error {
	data, err := format.Source([]byte(exp.Expression.Value))
	if err != nil {
		return writeIndent(w, indent, exp.Expression.Value)
	}

	_, err = w.Write(data)
	return err
}
```
If that's it, then you were right, `writeIndent` is more need if we successfully format with the builtin formater !

It's not necessarily obligatory I would say to make this change but I'm still going to make a commit since theoretically it's a performance gain. 

(At least I think so? Afterwards it's not necessarily very important to be a little slower or fast)
​

jtarchie:
I was merely making an observation. I defer to @a-h on what they would want.

gungun974:
> I was merely making an observation. I defer to @a-h on what they would want.

It may have just been an opinion but I actually think it's a good idea. 

The reason I started writing Go was to write simple code in a simple language.

So it seems simpler and faster to me to do a Write directly than to have to apply a function that no longer makes sense.

In addition I want to add but unit tests are also designed to test and refactor things that we didn't think about at the beginning ^^

gungun974:
In any case I'm just going to ask a quick question here on Github since I'm not yet used to contributing that much.

For the small modification I did not want to create an additional commit because it was a fix of a commit which is not part of the main branch. (And also I don't know what message I could write)

So I did a `git rebase` to modify the commit.

This changed the IDs so my commits created everything new. 

To put it on Github I had to `git push force` since I altered the history.

My question is does this kind of thing make sense in a PR or not?
Is it a bad thing ?
Is it more important in a PR to have a clean list of modification with multiple commit or have multiple commit that represent the evolution of the suggestion ? 

In any case I'm just going to ask a quick question here on Github since I'm not yet used to contributing that much.

For the small modification I did not want to create an additional commit because it was a fix of a commit which is not part of the main branch. (And also I don't know what message I could write)

So I did a `git rebase` to modify the commit.

This changed the IDs so my commits created everything new. 

To put it on Github I had to `git push force` since I altered the history.

----

My question is does this kind of thing make sense in a PR or not?
Is it a bad thing ?
Is it more important in a PR to have a clean list of modification with multiple commit or have multiple commit that represent the evolution of the suggestion ? 

I don't think there is one way to think about that and both are valid but I never have the ocasion to just ask.

a-h:
In a PR, you can make as many commits as you like. I find it helpful to see the evolution of the PR.

However, when I merge the PR, I'll "squash" them into a single commit using a "conventional commit" message so that only one single commit is made to the main branch. The commit message history helps to form the release notes, hence standardising them.

* https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits
* https://www.conventionalcommits.org/en/v1.0.0/

a-h (approved):

a-h:
Thanks, that looks great. I've tested it with the LSP and didn't see any issues there.

PR #214, opened 2023-10-05
fix: hang after m.Lock in KillAll (caused by #181)

headblockhead:
My recent PR (#181) had an issue where it would hang on the run on KillAll, as it would lock the mutex, then attempt to lock it again. This has been resolved by moving the KillAll to before the mutex is locked as part of Run.

This PR also renames KillAll to Stop for clarity.


a-h (approved):

a-h:
Thanks for that! 😁

Issue #213, opened 2023-10-05
`templ fmt` broke my HTML text


gungun974:
So my problem is very annoying.

When I render with `templ` with a simple HTML Paragraph like the example below :

```html
<p class="my-4 text-justify text-2xl font-light leading-9 text-text-dark md:my-16">
        Rencontrez l'équipe de La Barque, le cœur de notre restaurant.
        Dirigé par <strong> Gérard</strong>, notre équipe se compose du
        <strong> Chef Cédric</strong>, qui transforme les ingrédients frais
        en plats délicieux, et de <strong> Teddy</strong>, notre pâtissier,
        qui ajoute une touche sucrée à votre repas avec ses desserts faits
        maison. Chaque membre de notre équipe contribue à l'
        <strong>atmosphère chaleureuse et accueillante </strong>
        de La Barque, rendant chaque <strong> visite inoubliable</strong>.
</p>
```

I got this :

![2023-10-05_22-26](https://github.com/a-h/templ/assets/5224459/d11d85b0-2930-4ae5-99a4-17033df8bb1f)

Now if I use `templ fmt`. It change the HTML for that :

```html
<p class="my-4 text-2xl font-light leading-9 text-justify md:my-16 text-text-dark">
        Rencontrez l'équipe de La Barque, le cœur de notre restaurant.
        Dirigé par 
        <strong>Gérard</strong>
        , notre équipe se compose du
        <strong>Chef Cédric</strong>
        , qui transforme les ingrédients frais
        en plats délicieux, et de 
        <strong>Teddy</strong>
        , notre pâtissier,
        qui ajoute une touche sucrée à votre repas avec ses desserts faits
        maison. Chaque membre de notre équipe contribue à l'
        <strong>atmosphère chaleureuse et accueillante </strong>
        de La Barque, rendant chaque 
        <strong>visite inoubliable</strong>
        .
</p>
```
The problem now with this formatting is the HTML parsing of space is different and now I have some strange space between period and comma !

![2023-10-05_22-32](https://github.com/a-h/templ/assets/5224459/87575110-6312-4311-b15c-26ff105fc800)

I think `templ fmt` should not format HTML text and just keep what the user input. Even if it's long.
We can't realy know when to cut the text cause of this.

*(sorry if you don't understand French, I prefer show directly my issue with the text I need to use)*

gungun974:
So for information before anyone try to think about how to solve my problem.
I currently working on this issue with the formater.
[...]

PR #217, opened 2023-10-06
feat: make `templ fmt` respect user line break for elements like <span>

gungun974:
For more detail about why this change, look at my issue #213.

For the changes I made I introduce in the `Element` and `Text` Nodes a new property call `LineBreak`.

`LineBreak` can be true in two case while parsing.
1. When the user force the line break to make text read easy
2. When the user must break the line for element that are not sensitive to space like `<a>`. (Like the current behavior of the formater).

Next now I known in the Node Tree which Node need to be line break.

When we write those node in `writeNodes`, we prevent the line break if the node `LineBreak` is false.
However, that’s not all I need to do. I need to find if the previous block was on an other line or not cause If I don't. I will put the indent space in the middle of a line.

I think I can maybe improve the part when I check of indent but I can't succeed to scan backward in the text parser and element parser to get If I'm at the start of the line or not.

 In any case I hope that these changes are simple enough because I took some time to find the right formula and now when I look at my code I am no longer sure if without my new knowledge about Templ. I could understand a thing. :eyes:
*(But it's maybe just me)*


gungun974:
Sorry if Github closed this Pull Request, it's just that my branch had a conflict in a unit test and I had trouble using the Github code editor to resolve this.

So I discard my commit, refresh my branch, re apply my commit with a cherry pick and resolve in local the problem.

The only issue was for a moment my branch had no commit so Github automatically closes it...

parser/v2/elementparser.go
```diff
@@ -411,6 +411,23 @@ var element elementParser
...
+	"em",
+	"span",
+	"u",
+	"br",
+	"abbr",
+	"code",
+	"mark",
+	"small",
+	"q",
+}
```
a-h, line 429: My first thought is that this seems to be a list of inline elements and that the intent is that inline elements shouldn't be placed on a new line, unless the previous node is a text node, and ends with whitespce (since in this case, we're free to add more space, because it will be collapsed to a single space due to whitespace normalisation rules).

https://www.w3schools.com/html/html_blocks.asp

My second thought is that maybe it would be faster to use a dictionary lookup instead of iterating through the loop of elements.

Something like this, but for inline:

```
var blockElements = map[string]struct{}{
	"address": {}, "article": {}, "aside": {}, "body": {}, "blockquote": {}, "canvas": {}, "dd": {}, "div": {}, "dl": {}, "dt": {}, "fieldset": {}, "figcaption": {}, "figure": {}, "footer": {}, "form": {}, "h1": {}, "h2": {}, "h3": {}, "h4": {}, "h5": {}, "h6": {}, "head": {}, "header": {}, "hr": {}, "html": {}, "li": {}, "main": {}, "meta": {}, "nav": {}, "noscript": {}, "ol": {}, "p": {}, "pre": {}, "script": {}, "section": {}, "table": {}, "template": {}, "tfoot": {}, "turbo-stream": {}, "ul": {}, "video": {},
}

func (e Element) isBlockElement() bool {
	_, ok := blockElements[e.Name]
	return ok
}
```

parser/v2/elementparser.go
```diff
@@ -422,5 +439,18 @@ func (elementParser) Parse(pi *parse.Input) (r Element, ok bool, err error) {
 		err = parse.Error(fmt.Sprintf("<%s>: %s", r.Name, strings.Join(msgs, ", ")), start)
 	}
 
+	r.LineBreak = true
```
a-h, line 442: It's not easy to understand what "LineBreak" means. Does it mean that the element should not be placed on a newline because it's inline and the previous node is text and ends with whitespace?

parser/v2/elementparser.go
```diff
@@ -422,5 +439,18 @@ func (elementParser) Parse(pi *parse.Input) (r Element, ok bool, err error) {
...
+	for _, t := range ignoreNewLinesTags {
+		if r.Name == t {
+			r.LineBreak = false
+			break
+		}
+	}
+
+	if line, _ := pi.Peek(1); line == "\n" {
+		r.LineBreak = true
+	}
```
a-h, line 453: This might be easier to understand if it was a function.

I think the function might be called "shouldElementHaveATrailingLineBreak"?

```suggestion
	r.LineBreak = true
	for _, t := range ignoreNewLinesTags {
		if r.Name == t {
			r.LineBreak = false
			break
		}
	}
	if line, _ := pi.Peek(1); line == "\n" {
		r.LineBreak = true
	}
```

parser/v2/forexpressionparser_test.go
```diff
@@ -57,6 +57,7 @@ func TestForExpressionParser(t *testing.T) {
 								},
 							},
 						},
+						LineBreak: true,
```
a-h, line 60: This change needs to be made in a lot of places. Would it make more sense to invert it, so that it's "NoTrailingLineBreak"? That way, it would be false by default.

parser/v2/textparser.go
```diff
@@ -18,5 +20,22 @@ var textParser = parse.Func(func(pi *parse.Input) (t Text, ok bool, err error) {
...
+			if strings.Contains(v, "<"+i+">") || strings.Contains(v, "</"+i+">") {
+				t.LineBreak = false
+				break
+			}
+		}
+	}
+
+	if line, _ := pi.Peek(1); line == "\n" {
+		t.LineBreak = true
+	}
```
a-h, line 38: I'm not convinced by this section.

This looks like it's peeking forward into potentially non text code to look for elements.

a-h (changes requested):
I've left a question here about what I think the intent of the PR is.

I *think* that what you're saying is that inline elements are put onto a new line. So that this HTML...

```html
<div>
  This is <b>some text</b> in a sentence. <b>T</b>he first character of this sentence is bold.
</div>
```

Becomes this.

```html
<div>
  This is 
  <b>some text</b>
  in a sentence. 
  <b>T</b>
  he first character of this sentence is bold.
</div>
```

This is very wrong, because it adds a space character between `<b>T</b>` and `he`.

Is that about right?

a-h:
Oh, and thanks for looking into this!

gungun974:
> I've left a question here about what I think the intent of the PR is.
> 
> I _think_ that what you're saying is that inline elements are put onto a new line. So that this HTML...
> 
> ```
> <div>
>   This is <b>some text</b> in a sentence. <b>T</b>he first character of this sentence is bold.
> </div>
> ```
> 
> Becomes this.
> 
> ```
> <div>
>   This is 
>   <b>some text</b>
>   in a sentence. 
>   <b>T</b>
>   he first character of this sentence is bold.
> </div>
> ```
> 
> This is very wrong, because it adds a space character between `<b>T</b>` and `he`.
> 
> Is that about right?

That exactly what this PR change and what is the problem describe in issue #213.

I will look at your comments now, respond to some of them and make some change of course ^^

parser/v2/elementparser.go
gungun974: Oh thank you I didn't know that on w3schools there was a page clearly given Inline and Block Elements. I will so change the list to been accord with inline element cause they are the only that must subject to this change.

For the dictionary thing. I didn't think first cause for me Dictionary is for key value pair but that clever !
I will use that rather then an array ^^

parser/v2/elementparser.go
gungun974: "LineBreak" is to indicate that we must put a line break after the element.

By default we assume that every element should line break except our Inline Element.

But let me explore more `IndentChildren` after re locking your code here I have some feeling they was already something with this behavior I could just change.

parser/v2/forexpressionparser_test.go
gungun974: Personally I never wrote an property with a negative sign but since In go I can't say that Element is by default true. That better.

After maybe I can get rid of this `LineBreak` property

parser/v2/textparser.go
gungun974: Yes by default the parser assume that every line after text should got a new line.

So here I have try some rule to decide if this text must wrap or not after.

But now I re read my code I see some behavior that are useless.

Text should always wrap except if after it there is an inline tag.

This rule can just be made in the `types.go` file when we reconstruct the code.

I will try so the optimize this and so try to just don't touch this file.

gungun974:
So I completely rework my PR with your suggestion and the new idea I had.

Like I said, I found multiple thing strange I was doing when exploring what I can do but now I think ***I found a nicer solution for the problem***.

### Here a summary of my change compared to your comments :

- I follow your advice for the dictionary and Implement on `Element` the method `isBlockElement`.

- For my new solution I need to know if an element is at the end or not of a line. 
  For that I introduce `NotAtEndOfLine` since It was the least current.

- I didn't made `shouldElementHaveATrailingLineBreak` cause my code in `elementparser.go` now only check if we are at the end of line or not and I think it's self-explanatory (but we can debate on this).

- **The ugly thing I made `textparser.go` don't exist anymore.**

### Now here my new approach :

`writeNodes` is the master function that put our element.
It's has the responsible to make the line break and indent.

So my new approach consist to give `writeNodes` more control about line break and indent with `isStartOfTheLine` and `ignoreLineBreak`.
`isStartOfTheLine` is a rule that make sure we always make the indent at the start of the line.
`ignoreLineBreak` stop the behavior of breaking line.
- We should ignore the breaking line only if the element is an inlineElement except when this element is at the end of a line
- Text should always break the line except if the next element is an inline element.

With all those rule we can easily control how indent work.

Maybe there are some area to improve in this solution but but we are still closer to the final version ^^

parser/v2/types.go
```diff
@@ -173,7 +173,7 @@ type Whitespace struct {
 func (ws Whitespace) IsNode() bool { return true }
 
 func (ws Whitespace) Write(w io.Writer, indent int) error {
-	if ws.Value == "" || !strings.Contains(ws.Value, "\n") {
+	if ws.Value == "" || strings.Contains(ws.Value, "\n") {
```
gungun974, line 176: How yes and also I edit those lines cause I think there was an error and this error make my test don't work.
Basically `writeNodes` would sometime write a space cause in the original text we had space. 
```html
<strong>Souffle</strong> chocolate
```
The issue was that condition block the write of the tiny whitespace cause that tiny whitespace don't have a line line.
But since it's always `writeNodes` that is responsible for new line.
I think this is an error cause we should never write new line everywhere except for `writeNodes`.

gungun974:
This PR is no needed anymore since #224 has a better way to do and some extra thing.

PR #224, opened 2023-10-09
fix: issue 213

a-h:
Rework the parser, and generator to be able to format inline elements within templ files, but also strip unnecessary whitespace from rendered output.

@gungun974, can you check this out please? I spent a while trying to understand it all in detail, and this is what I ended up with.

@joerdav - could use another pair of eyes on this if you can! 😁


gungun974:
Hmm nice job @a-h ! 

Basically, if I had done PR #217 it was to quickly set up a functional formatter for my personal project without forcing your hand but seeing that you had taken the time to work on this in your own way. That's very nice !

So before talking about this PR. I can confirm this PR has the same behavior of my version at #217. I took your formatter and my version and parse on the example of #213 and both yield the *same result. 
*(In reality there was a difference of the ending line of the file that was remove but who care ?)

Now about that PR. I'm not an expert in your project compare to you but for what I can say. This look nice and clean.

----

However I don't understand the test `all children indented, with nested indentation, when close tag is on same line`

Should the result not be this ? :
```templ
templ input(value, validation string) {
	<div>
		<p>
			{ "the" }<a href="http://example.com">{ "data" } </a>
		</p>
	</div>
}
```
----

And just a last nick-pick. I look on prettier to see how prettier mange a HTML `<br>`. 
In prettier `<br>` automatically break the line after the element.
Your implementation break the line before.
Okay nobody care but it's a thing I wanted to said cause why not.

----

Anyway thank again for this PR cause it's of course better than #217 and I would rather it be the creator of this magnificent parser who perfects it than me with code that looks more like a bad hack.
(I'm hard on myself I know but hey I really mean it ^^)

joerdav:
This makes sense to me, I can see a couple of test templates got updated, were they expected changes based on this change? They seem correct to me

a-h:
@gungun974 - on test you didn't understand the output of, I think it's because there wasn't a clear description of it.

These test name better describe the expected behaviour.

* non-empty elements whith children that are all on the same line are not split into multiple lines
* when an element contains children that are on new lines, the children are indented

```go
{
	name: "non-empty elements whith children that are all on the same line are not split into multiple lines",
	input: ` // first line removed to make indentation clear in Go code
package test

templ input(value, validation string) {
	<div><div><b>Text</b></div></div>
}
`,
	expected: `// first line removed to make indentation clear in Go code
package test

templ input(value, validation string) {
	<div><div><b>Text</b></div></div>
}
`,
},
{
	name: "when an element contains children that are on new lines, the children are indented",
	input: ` // first line removed to make indentation clear in Go code
package test

templ input(value, validation string) {
	<div>
	<div><b>Text</b></div></div>
}
`,
	expected: `// first line removed to make indentation clear in Go code
package test

templ input(value, validation string) {
	<div>
		<div><b>Text</b></div>
	</div>
}
`,
},
```

On the `br`, I've added an extra rule for that.

So... I think that solves the nits.

@joerdav - yes, the rules have changed on whitespace handling. It all used to get stripped, but there are some cases where it needs to remain (e.g. between `{ "string" } { "string" }` elements, or `<b>D</b>on't lose or introduce <b>spacing</b> around inline nodes."

Issue #229, opened 2023-10-12
Simplifying the use of Templ components with HTMX


gungun974:
I was working on an HTMX project with Templ and I was wondering if there was a simpler way to use Templ components with HTMX.

In my application, to avoid reinventing my button everywhere, I created a Component like this :

```templ
type AppButtonParams struct {
  Variant string
  Size    string
  Class   string
}

templ AppButton(params AppButtonParams) {
  <button
    class={ 
      templ.SafeClass(appButtonVariants(params.Variant, params.Size)),
      templ.SafeClass(params.Class),
    }
  >
    { children... }
  </button>
}
```

The challenge now is that I want to use an `hx-get` attribute for a specific button. With the current syntax of these components, I can't transparently pass new attributes to my `AppButton` component since it's a function.

Is the only way to pass props to a Templ component by passing values into the function, or is there a syntax sugar I can use to simplify this code ?

```templ
type HTMXParams struct {
  Get    string
  Target string
  Swap   string
}

type AppButtonParams struct {
  Variant string
  Size    string
  Class   string
  HTMX    HTMXParams
}

templ AppButton(params AppButtonParams) {
  <button
    class={ 
      templ.SafeClass(appButtonVariants(params.Variant, params.Size)),
      templ.SafeClass(params.Class),
    }
    if !helpers.IsEmptyOrWhitespace(params.HTMX.Get) {
      hx-get={ params.HTMX.Get }
    }
    if !helpers.IsEmptyOrWhitespace(params.HTMX.Target) {
      hx-target={ params.HTMX.Target }
    }
    if !helpers.IsEmptyOrWhitespace(params.HTMX.Swap) {
      hx-swap={ params.HTMX.Swap }
    }
  >
    { children... }
  </button>
}
```

stefanerickson:
Hey @gungun974, I'm just getting to know Templ, but seems you could separate out the attribute rendering and define your own types for HTML params... Haven't tested this, but would it work for to do the following?:

```go
type HTMXParams struct {
  Get    string
  Target string
  Swap   string
}

type AppButtonParams struct {
  Variant string
  Size    string
  Class   string
  HTMXParams           // Embedded
}

func renderAttributes(attrs map[string]string) string {
  var result []string
  for key, val := range attrs {
    if val != "" {
      result = append(result, fmt.Sprintf("%s=\"%s\"", key, val))
    }
  }
  return strings.Join(result, " ")
}

templ AppButton(params AppButtonParams) {
  <button
[...]

PR #237, opened 2023-10-17
Spread attributes

gungun974:
### Introduction

This Pull Request was created to address the primary issue of #229.

The Templ component system prevents passing attributes directly to an element within the component.

This can be quite problematic as it complicates the use of libraries like HTMX, alpinejs, or hyperscript due to an almost infinite number of attributes.

This Pull Request adds a "spread" attribute behavior to Templ similar to what can be found in React.

### The "spread attributes" syntax

**"Spread attributes"** are a new type of attribute that allows adding directly to the generated code a `map[string]string` as a list of an element's attributes.

For this, the Templ generator detects the "spread attributes" syntax and adds to the code a `for range` of this map and escapes the name and the value of the attribute at runtime.

An alias `templ.Attributes` is available in the runtime to enhance visibility and avoid writing `map[string]string`.

Here's an example of code written with this syntax.

```templ
package spread_attributes

templ AppButton(attrs templ.Attributes) {
  <button { attrs... } />
}

templ Consume() {
  @AppButton(templ.Attributes{"data-id": "my-custom-id"}) {
    Click
  }
}
```

This syntax was inspired by the proposal from @joerdav.
### Todo

This Pull Request is not yet complete; here are the missing points:

- [x] Recognize the "spread attributes" syntax
- [x] Generate runtime code handling the "spread attributes"
- [x] Update the documentation to explain the operation of "spread attributes"
~~[ ] (Not sure) Overwrite conflicting attributes (This could have a higher runtime impact as this would break constant attributes with the current operation)~~ (It's not my responsibility in the end)


xgalaxy:
I'm not as familiar with the code as you are. But you might want to make sure this doesn't clash with `children...`

It looks like the parser specifically looks for `children...` so there might not be an issue. I don't know if templ throws a parsing error if a variable coming in contains the name `children` or not so maybe a potential clash is already impossible?

gungun974:
> I'm not as familiar with the code as you are. But you might want to make sure this doesn't clash with `children...`
> 
> It looks like the parser specifically looks for `children...` so there might not be an issue. I don't know if templ throws a parsing error if a variable coming in contains the name `children` or not so maybe a potential clash is already impossible?

This is not an issue, the parser is divided into smaller parsers, and each smaller parser has a higher priority than the other. 

So here we have the `element.Parse` parser which is executed before `childrenExpression.Parse`. 

So there isn't really any conflict. Moreover, it's important to understand that the `{ children... }` for injecting a child is not real. 
If you look at the generated template, you won't actually see a "children" variable. 
What the generator does is that it generates code that fetches the child from the runtime context and calls the render method. 

However, for safety, in case the order changes one day, I will modify my unit tests. ^^

docs/docs/03-syntax-and-usage/03-attributes.md
```diff
@@ -63,6 +63,30 @@ templ component() {
 <hr style="padding: 10px" class="itIsTrue" />
 ```
 
+## Spread attributes
+
+Using `{ spread... }` in an element. You can easily transfer attributes from a `map[string]string` into your elements.
```
joerdav, line 68:
```suggestion
Use the `{ attrMap... }` syntax in the open tag of an element to append a dynamic map of attributes, where `attrMap` is of type `templ.Attributes`.

`templ.Attributes` is a `map[string]string` type definition.
```

docs/docs/03-syntax-and-usage/03-attributes.md
```diff
@@ -63,6 +63,30 @@ templ component() {
...
+
+To simplify, you can use `templ.Attributes{}` instead of `map[string]string`.
```
joerdav, line 70:
```suggestion
```

parser/v2/elementparser.go
```diff
@@ -259,6 +259,52 @@ var expressionAttributeParser = parse.Func(func(pi *parse.Input) (attr Expressio
...
+	func(pi *parse.Input) (attr SpreadAttributes, ok bool, err error) {
+		start := pi.Index()
+
+		// Optional whitespace leader.
+		if _, ok, err = parse.OptionalWhitespace.Parse(pi); err != nil || !ok {
+			return
+		}
+
+		// {
+		if _, ok, err = parse.Or(parse.String("{ "), parse.String("{")).Parse(pi); err != nil ||
```
joerdav, line 272: Could use `openBraceWithOptionalPadding` parser here.

parser/v2/elementparser.go
```diff
@@ -259,6 +259,52 @@ var expressionAttributeParser = parse.Func(func(pi *parse.Input) (attr Expressio
...
+			!ok {
+			pi.Seek(start)
+			return
+		}
+
+		// Expression.
+		if attr.Expression, ok, err = exp.Parse(pi); err != nil || !ok {
```
joerdav, line 279: Potential use case for a `parser.Until`, will prevent you from having to trim as below.

joerdav: It may not work in this case actually. But worth thinking about!

generator/test-spread-attributes/template.templ
```diff
@@ -0,0 +1,7 @@
+package testspreadattributes
+
+templ BasicTemplate(spread templ.Attributes) {
```
joerdav, line 3: Could be worth a test that combines the spread with a conditional, just to prove that it works.

docs/docs/03-syntax-and-usage/03-attributes.md
```diff
@@ -63,6 +63,30 @@ templ component() {
...
+  @component(templ.Attributes{"data-testid": "paragraph"}) 
+}
+```
+
+```html title="Output"
+<p data-testid="paragraph">Text</p>
+```
+
+:::note
+You can use conditional attributes with the spread attributes.
```
joerdav, line 87: Rather than a note here, what about an example in the conditional attributes section that shows it in used?

joerdav (commented):
Awesome PR!

docs/docs/03-syntax-and-usage/03-attributes.md
gungun974: Something like this ?

image

Personally, I find that the example is a lot for a niche case.

joerdav:
It looks like due to `map` semantics, the attributes are appearing in a random order. Either need to loop through in an ordered way or change to a slice 
```go
type Attribute struct {
	Key, Value string
}
type Attributes []Attribute
```

The usage wouldn't be too different if it was moved to a slice:

```
	component := BasicTemplate(templ.Attributes{
		{"id",     "test"},
		{"hx-get", "/page"},
	})
```

docs/docs/03-syntax-and-usage/03-attributes.md
joerdav: I don't think it needs a whole example in it's own. I thought you could just amend the example in https://templ.guide/syntax-and-usage/attributes#conditional-attributes to contain a spread?

gungun974:
> It looks like due to `map` semantics, the attributes are appearing in a random order. Either need to loop through in an ordered way or change to a slice
> 
> ```go
> type Attribute struct {
> 	Key, Value string
> }
> type Attributes []Attribute
> ```
> 
> The usage wouldn't be too different if it was moved to a slice:
> 
> ```
> 	component := BasicTemplate(templ.Attributes{
> 		{"id",     "test"},
> 		{"hx-get", "/page"},
> 	})
> ```

Yes, I was going to say that I had forgotten that Go randomly randomizes the maps and therefore that it breaks the tests.

So in HTML it's not a problem but it's annoying.

I personally think it would be better to don't change the signature and just make the runtime insert attribute by alphabetical order.

gungun974:
I had already written this for a personal project:
```go
sortedAttributes := make([]string, 0, len(attributes))

for k := range attributes {
    sortedAttributes = append(sortedAttributes, k)
}

sort.Strings(sortedAttributes)

for _, k := range sortedAttributes {
    v := attributes[k]
}
```
I don't think this will have such a negative impact on performance and at least we are sure of the order.

gungun974:
> It looks like due to `map` semantics, the attributes are appearing in a random order. Either need to loop through in an ordered way or change to a slice
> 
> ```go
> type Attribute struct {
> 	Key, Value string
> }
> type Attributes []Attribute
> ```
> 
> The usage wouldn't be too different if it was moved to a slice:
> 
> ```
> 	component := BasicTemplate(templ.Attributes{
> 		{"id",     "test"},
> 		{"hx-get", "/page"},
> 	})
> ```

Also now I think using slice could make some runtime mistake if someone create a slice with only the key and not a value or strange case when someone create an attribute with a key, a value and something else.
```go
component := BasicTemplate(templ.Attributes{
  {"id"},
  {"hx-get", "/page", "third"},
})
```

joerdav:
@gungun974 That would not be the case:

https://go.dev/play/p/-wX7DtB9JrN

joerdav:
Yep, don't think performance would be too much of an issue either, these maps should be relatively small.

gungun974:
> @gungun974 That would not be the case:
> 
> https://go.dev/play/p/-wX7DtB9JrN

Oh you make me learn something.
I still prefer sort map since map got some extra method link to them but yes, it's not all that bad.

parser/v2/elementparser.go
gungun974: > Potential use case for a `parser.Until`, will prevent you from having to trim as below.

This can't work here because the expression parser grab the three dots.

The only thing I can do is try to change the "expression parser" but the issue is if I remove the ellipses from the value.

I can't after in the `spreadAttributesParser` check if In my expression I got my ellipses since I need to look into the past.

I think for now the simplest thing is as it its now. *(Note, I try things before saying that)*

docs/docs/03-syntax-and-usage/03-attributes.md
gungun974: > I don't think it needs a whole example in it's own. I thought you could just amend the example in https://templ.guide/syntax-and-usage/attributes#conditional-attributes to contain a spread?

Like this ?

image

I admit I find it cleaner

joerdav (commented):
This all looks good, will need @a-h to take a look and decide if this is the API we want to go with.

My only reservation is that the `...` syntax in Go is for slices/arrays rather than maps, but maybe that is okay.

gungun974:
> My only reservation is that the `...` syntax in Go is for slices/arrays rather than maps, but maybe that is okay.

I can understand `...` can be strange since we give here an another meaning.

But to argue with this syntax. Templ already use `children...` to insert child element and those ellipse or not really common in go (there are not spread operator used everywhere like JS :skull:).

---

> This all looks good, will need @a-h to take a look and decide if this is the API we want to go with.

If you (@joerdav) or @a-h want to change the way the parser represent that new attribute type or the way the generator generate.

I'm willing of course to improve this PR ^^

skydiver1980 (approved):

a-h:
Sorry for not merging this. Still not sure I've thought through all of the consequences, and want to get the design right.

I was thinking about how you would send things like optional attributes down the chain.

There are also boolean attributes that don't have a key/value pair, e.g. `noshade`. They just appear in HTML.

I thought the design might need to be more than a `map[string]string`, and be a `map[string]Attribute` to accommodate that.

```templ
package spread_attributes

type Attribute struct {
  Name string
  Value string
  OmitValue bool // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
  Visible func(ctx context.Context) bool
}

templ AppButton(attrs templ.Attributes) {
  <button { attrs... } />
}

type AttrOpt func(*templ.Attributes)

func ConditionalAttr(condition bool, name, value string) AttrOpt {
  return func(a *templ.Attributes) {
      a[name] = value
      a[name].Visible = func(ctx context.Context) bool { return condition }
  }
}

func Attr(name, value string) AttrOpt {
  return func(a *templ.Attributes) {
    a[name] = value
  }
}

func BooleanAttr(name string) AttrOpt {
  return func(a *templ.Attributes) {
    a[name].OmitValue = true
  }
}

func Attributes(opts ...AttrOpt) templ.Attributes {
  a := make(templ.Attributes)
  for _, opt := range opts {
    opt(a)
  }
  return a
}

templ Consume() {
  @AppButton(Attributes(BooleanAttr("noshade"), ConditionalAttr(isLoggedIn, "class", "authenticated"), Attr("data-id": "my-custom-id"))) {
    Click
  }
}
```

Or am I overthinking this one?

gungun974:
So it's complicated for me to say if you "overthink" since I'm biased about my syntax but looking with the spec of HTML boolean attributes you have shown.

> https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

 It's written : `If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.`

That mean in HTML you can write a `disabled` attribute with an empty value or the `disabled` value. In fact I will don't lie to you but I never write the shorten version without the value been the same :sweat_smile:

So for the example of `noshade` nothing prevent you to use the longer syntax. Of course if Templ try to aim to the less HTML generate it's more verbose but I think HTML is cheap with gzip so 6 characters more is not the end of the world.

---

Otherwise I think this Syntax can be Interesting. I think with some formatting it can be regular to read
```templ
templ Consume() {
  @AppButton(
    Attributes(
      BooleanAttr("noshade"),
      ConditionalAttr(isLoggedIn, "class", "authenticated"),
      Attr("data-id": "my-custom-id"),
    )
  ) {
    Click
  }
}
```
My only issue is that make the user need to learn a lot of custom structure just to do at the end a spread of attributes :eyes:

When you said to someone it's just a `map[string]string` it's easy to understand, as a smaller footprint when writing and since it's just a well know type in go, You know how write function to manipulate. 

The only thing this I see in this Syntax that is better from my current PR is `ConditionalAttr`. With a map you can't use ternary (in fact we don't have ternary).

This can be very annoying since you have to create a helper function returning the write value. But I think this feature can be implemented if Templ has a way to define go code directly in the template. You could just define early the map, do the if append and then use it.

*I think you (@a-h) mention early in an other Issue you planed to add this native `go code` injection in Templ. This feature can really help to simplify the user and the codebase of this project itself since it's give to the develop the power of writing code that solve they specific problem than trying to create some helper or complex structure like here.*

Anyway that my take on this problem and I could be very wrong but I think it's the simpler and more go user friendly we can do for now.

cmar0027:
> Or am I overthinking this one?

Sorry for the intrusion, but what about  just having `type templ.Attributes = map[string]interface{}`, then generate code that uses type assertion like this:
- if the value is a string just add the attribute with its value
- if the value is a bool, add the attribute as a boolean attribute only if value is true 
- if the value is a templ.KeyValue[string, bool], this would be an optional attribute

```
templ AppButton(attrs templ.Attributes) {
  <button { attrs... } />
}

templ Consume() {
  @AppButton(templ.Attributes{
    "nonshade": true,
    "class": templ.KV("authenticated", isLoggedIn),
    "dateId": "my-custom-id",
  }) {
    Click
  }
}

```

I don't know the templ code base to say what approach would be faster, but this declarative looks more intuitive to me.

gungun974:
> > Or am I overthinking this one?
> 
> Sorry for the intrusion, but what about just having `type templ.Attributes = map[string]interface{}`, then generate code that uses type assertion like this:
> 
>     * if the value is a string just add the attribute with its value
> 
>     * if the value is a bool, add the attribute as a boolean attribute only if value is true
> 
>     * if the value is a templ.KeyValue[string, bool], this would be an optional attribute
> 
> 
> ```
> templ AppButton(attrs templ.Attributes) {
>   <button { attrs... } />
> }
> 
> templ Consume() {
>   @AppButton(templ.Attributes{
>     "nonshade": true,
>     "class": templ.KV("authenticated", isLoggedIn),
>     "dateId": "my-custom-id",
>   }) {
>     Click
>   }
> }
> ```

Oh that's a great Idea I love it ! :heart:

That would mean the developer can pass everything and in the runtime we only easily ignore the attribute if the value is not valid (for example if it's was a float). 

Also the fact we reused `templ.KV` for conditional is Neat ! 

> I don't know the templ code base to say what approach would be faster, but this declarative looks more intuitive to me.

Of course the loop will be slower since we need to cast if the `interface{}` is a `string`, `boolean` or `templ.KV` but I don't think it's the kind of operation is very that slow.

The thing that hype me the most it's the kind of little change you suggest @cmar0027 that's so clean and so perfect that we couldn't see at first gland with our eyes so deep in complex solution !

filipweidemann:
Hey, quickly chiming in: 

I would double down on the point that @gungun974 made. We can easily write attribute values on e.g. `hidden` or others. 

Also, @cmar0027's syntax is great imho! With this, there is no need to learn all the new attribute structs.

gungun974:
Hum CI/CD failed, I guess I will need to rebase this branch and verify everything still work.

For now 75ea299ceae5c3a6aa8c4f692e38b18eb7c652b6 add what we have discuss yesterday.

I didn't have yet update the documentation since I'm too bad to write english and explain thing.

I will fix later this day CI/CD but I hope the new code I wrote will be at your standard @a-h

a-h:
Don't worry, I'll check out what's going on with the linter.

imaai:
@a-h @gungun974 Any updates on this one?

a-h:
OK, I've reworked this PR and I think the end result is a bit simpler!

a-h (approved):

filipweidemann:
You have no idea how happy I am right now! 
It really was a showstopped for me when trying to migrate an existing JS-based app to Go & templ, so with this PR I have all the (obvious) bits I need to pull this off. 

Thank you so much for your efforts @a-h, @gungun974 & everyone who contributed by either reviewing or shaping the feature otherwise!

the-goodies:
Am I misunderstanding something or you can't pass a single attribute and use it like: { attr }?
What is the point of templ.Attributes being map[string]any instead of []Attribute where Attribute is struct{Key string, Value any}? If it was a struct, you could pass a single value or pass a slice and use spread operation which works on slices in Go, unlike map which doesn't. This further adds while minor, but unnecessary differences between templ and Go for no apparent reason.

filipweidemann:
Hey, I'm no maintainer or even contributor (yet), but I want to quickly ask how the current implementation prevents you from achieving your goal here?

I am not even sure if having `[]Attr` would solve any real and existing issue. Is there something preventing you from using `templ.Attributes{"my-key", "my-value"}` & `{ attrs... }`? Is there a specific use case you currently cannot implement with this feature in its current state? You should absolutely be able pass single attributes using this syntax.

As for the syntax decisions: the feature is consistent with existing templ functionality, e.g. what @a-h  calls [child spread expressions](https://github.com/a-h/templ/blob/5d4b87f52dd76788732a6b6de9a03baa550ea9c7/parser/v2/goexpression/parse.go#L116)

I would argue that consistency inside the library itself is more important than parity with the underlying language. 
People use JSX after all. I would not call that HTML, nor JS. It is an extensions for JS for templating (primarily) React components (just like templ is for templating HTML using Go) & gets compiled to proper code. Same story if you'd ask me.

PR #238, opened 2023-10-17
feat(generator): support templ.URL with form action

aranw:
Noticed that it was not possible to use `templ.URL` with a form action


a-h (approved):
Thanks!

PR #241, opened 2023-10-19
Update flake and docs

angaz:
I've added some extra functionality to the flake to make it easier to install on NixOS, and to use within a flake project.

I also updated the docs to help someone new to flakes to include this project in their own project.

There's also a new package `templ-docs`, so if you are hosting the docs on a NixOS system, you can host the static files using `services.nginx`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,9 +7,13 @@
       url = "github:hercules-ci/gitignore.nix";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+    xc = {
+      url = "github:joerdav/xc";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
   };
 
-  outputs = { self, nixpkgs, gitignore }:
+  outputs = { self, nixpkgs, gitignore, xc }:
     let
       # Systems supported
       allSystems = [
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -2,36 +2,82 @@
...
+            npmDepsHash = "sha256-/9FzD38s0bER9zvR4AfU/KBrzRsWDmuFeuMgSL45/1s=";
+
+            installPhase = ''
+              mkdir -p $out/share
+              cp -r build/ $out/share/docs
+            '';
+          };
+        };
+
+        devshells.default = {
```
a-h, line 75: I'm not familiar with devshell and why I'd want to use it. Not sure why I'd take a library for this instead of using the built-in stuff?

As per above...

```nix
      devShell = forAllSystems
        ({ system, pkgs }:
          pkgs.mkShell {
            buildInputs = [
              pkgs.gotestsum # Alternative test runner.
              xc.packages.${system}.xc
              templ.packages.${system}.default
            ];
          });
```

a-h (changes requested):
Thanks a lot for the contribution.

The core of it is:

* Exporting an overlay.
* Improvements in the templ build step around setting the ldflags, CGO_ENABLED etc.
* Including a build for the docs.
* Including a dev shell with some tools.

On the overlay, I'm not convinced it's required, but I could be if there's a good reason. It's common to make your own anyway, e.g. https://github.com/headblockhead/dotfiles/blob/8e137383bd7bc7caf6350b2c67d92f4146e41f49/flake.nix#L62-L95

The rest is great, I'd take those, but would like to avoid the use of more libraries unless really required.

I'm happy to pick up the ideas and take it forward, or if you'd like to rework the PR, I'd also take that!

Just let me know.

flake.nix
angaz: I like devshell because it gives you some extra features like the ability to create commands, basically I use them as aliases to run other scripts.

It also loads much faster than the standard mkShell. I believe it has some extra caching. This is pretty important to me because I work on a lot of stuff and move around a lot, creating many terminals in the same project, so the dev shell loading quickly is big. But it is not needed, so we can revert to vanilla stuff.

angaz:
Sure, if you don't want anything unnecessary, then I will update the PR to keep it vanilla.

angaz:
I believe everything should be good to go.

angaz:
I didn't see `gotestsum` used anywhere in the project, so I didn't include it in the `devShell`. Tell me if you would like me to add it.

a-h (approved):
Looks great, thanks so much!

PR #243, opened 2023-10-20
Add support for formatting `ExpressionAttribute` with multiple line

### LINKED ISSUE #242: Templ formater don't format ExpressionAttribute with multiple line.
In Templ, a lot of time I need to write code like this :

```templ
<button
  class={ 
    templ.SafeClass("p-1.5 px-3 text-sm rounded"),
    templ.SafeClass("border border-two text-two hover:border-two-500 hover:text-two-300"),
  }
></button>
```

Code like this are totally valid and called in the documentation `Class expression`. 

My issue is the Templ parser see the class expression as an `ExpressionAttribute` with an `Expression` inside.

```go
{ 
    templ.SafeClass("p-1.5 px-3 text-sm rounded"),
    templ.SafeClass("border border-two text-two hover:border-two-500 hover:text-two-300"),
}
```

This is okay on the paper but the Templ formater don't do any formatting inside expression.
So at the end the formater don't really format those `Class expression`

```templ
<button
  class={ 
        templ.SafeClass("p-1.5 px-3 text-sm rounded"),
               templ.SafeClass("border border-two text-two hover:border-two-500 hover:text-two-300"),
      }
></button>
```

gungun974:
Resolve #242 

Here I edit the way expressionAttribute are stringify by passing their expression to `go/format` only when those expression are multiline and contain period (sign of of a `Class expression` / slice)

Normally `go/format` can't format non valid Go Code so since technically we are in a virtual slice, I make the code append then after format remove a real slice.

It's maybe a little bit hacky but normally `Expression` should always have some valid go code so in the best I world, I shouldn't have to do that.

As usual, I am always open to suggestions.

### DIFF
--- a/parser/v2/types.go
+++ b/parser/v2/types.go
@@ -98,6 +98,32 @@ type Expression struct {
 	Range Range
 }
 
+// Format go code in expression
+func (e *Expression) Format(formatLikeSlice bool) string {
+	if formatLikeSlice {
+
+		data, err := format.Source([]byte(`[]string{ ` + e.Value + ` }`))
+		if err != nil {
+			return e.Value
+		}
+
+		// Remove Start of slice
+		formatted := strings.Replace(string(data), "[]string{", "", 1)
+
+		// Remove end of slice
+		formatted = strings.Replace(formatted, "}", "", 1)
+
+		return formatted
+	}
+
+	data, err := format.Source([]byte(e.Value))
+	if err != nil {
+		return e.Value
+	}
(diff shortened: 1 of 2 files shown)


a-h:
Just wondering if we needed the format operation. The test can pass with a bit of string manipulation alone.

a-h:
Sorry for the long time to get to this by the way...

a-h:
Thanks a lot for the contribution!

PR #244, opened 2023-10-20
adds more standart html tags in snippets

BKR-dev:
hey, found that there are some html tags that would be nice to have around, i know that there is a "default empty" tag that can be used but i thought having more makes it nicer to get going.

hopefully you find this helpful or can give some feedback.

im only starting to work with templ and im amazed by the tooling around it and its LSP, im super excited! 

cheers!

### DIFF
--- a/cmd/templ/lspcmd/proxy/snippets.go
+++ b/cmd/templ/lspcmd/proxy/snippets.go
@@ -18,6 +18,12 @@ var htmlSnippets = []lsp.CompletionItem{
 		InsertTextFormat: lsp.InsertTextFormatSnippet,
 	},
 	{
+		Label:            "button",
+		InsertText:       `button type="button" ${1:}>${2:}</button>`,
+		Kind:             lsp.CompletionItemKind(lsp.CompletionItemKindSnippet),
+		InsertTextFormat: lsp.InsertTextFormatSnippet,
+	},
+	{
 		Label: "div",
 		InsertText: `div>
 	${0}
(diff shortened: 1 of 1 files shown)


a-h:
Looks good, but why so few? Do you think we should add more?

The built-in HTML support includes a larger set of tag snippets. I'm not sure if it's contextual, in that it's aware of the parent tag etc.

image

It seems I just added a placeholder for future thinking in this area. I'll merge for now, and we can revisit later.

BKR-dev:
Thanks for merging! 

>Looks good, but why so few? Do you think we should add more?

this where just he ones i thought of immediately when starting to work with it. im working on a PoC of my Org so im keeping the perspective of "a fresh new user without knowledge", which also means that in my setup i dont have additional lsp or plugin for html support, so i look at it "what can / should templ lsp provide" vs should not be. 

i work with a fairly small neovim setup and since i already have templ installed i naturally concluded i would use its lsp capability (which i love btw, so cool!)  to write the html part. so from that point i would think that there are tags "missing" but than the question is should the templ lsp part also provide complete html support, im not actually sure, maybe you have a thought about it? 

(i can see the argument for having all html tags in a separate lsp for html and therefore maybe providing htmx tags?)

BKR-dev:
Thanks for the feedback! 

Do I understand you correctly in the way that you want to incorporate additional html lsp and css lsp into the templ tooling? would that then mean to add htmx tags into the templ lsp?

I have joined the templ slack so we can discuss this on there when it's more about finding the way forward and not strictly tied to a merged PR. 

Thank for your time and input!

a-h:
Yes, I think it's possible to add more tooling into the VS Code extension or LSP by using the sample that Microsoft provide.

So I don't think there's a need to anything `htmx` specific to the LSP.

BKR-dev:
Im not sure i understand clearly what you mean. Do you mean to find a way to incorporate existing html and css lsp to Templ to provide that functionality? Im not sure there is an existing htmx lsp out there in the wild but adding that too would be nice as well. I saw your talk at gopherconUK yesterday and that helped to understand how the lsp works, very insightful! 

I would look into that as im working with it and find the need to have something like that working within my nvim setup.

a-h:
Yes, I think that there's a way to incorporate existing LSPs as per https://code.visualstudio.com/api/language-extensions/embedded-languages

Given that templ is basically a way to combine HTML + Go + CSS + JS, then I was wondering if the same approach I've used to proxy to the Go LSP would work, or whether the samples on that page ^ might be easier.

 Glad the Gophercon talk had the desired effect. 😁

BKR-dev:
Okay, i understand! Thanks for clarifying and taking the time to spell it out for me 😄

joerdav:
Also worth noting there is a htmx LSP under development already at github.com/ThePrimeagen/htmx-lsp

PR #248, opened 2023-10-22
added changes in readme command

bhargavshirin:


valxntine:
templ uses `xc` as it's task runner, using the `Directory:` tag before a command tells `xc` where to run the command in. 

You can read more about `xc` [here](www.github.com/joerdav/xc)

bhargavshirin:
sorry, but why I was marked spam. I didn't know about that

a-h:
Because your pattern of behaviour is to make minor changes that have no positive impact, or to run a formatting tool.

image

bhargavshirin:
i mainly focus on documentation. since its the main thing through which
people understand about the project/application/service. but i never
committed any unnecessary things. and coming to those commit messages i
just use the previously used commit messages for all the commits.

bhargavshirin:
@a-h could you please remove the spam label. since I have not made any invalid or unnecessary PR. you could even check out the other PRs which you have sent the image of.the commit message is similar but the commits are different.

PR #255, opened 2023-10-28
feat: proxy retries with exponential backoff

dvorakluk:
Proxy will retry on error for 10 times, so server has some time to come up after reloading/restarting.

Tested with `templ generate --watch --proxy="http://127.0.0.1:8089" --proxyport="8080" --cmd="go run ."`

Open to ideas whether the backoff should be user configurable.

based on #146


a-h (approved):
Thanks, this looks great!

PR #264, opened 2023-10-30
templ_internal package created

otaxhu:
Almost all of the API was moved from `templ` package to `templ_internal`. The respective corrections were made in the generator package.

## Reason:
We can all agree that the library was becoming cluttered with many functions and types that the consumer will almost never use (since almost everything will be generated by the templ CLI). So that is my solution to that problem, it cost me a lot to do this refactoring, any questions and edits would be wonderfully welcome.

You probably think I'm crazy since I modified more than 50 files and 1000 lines of code hahaha, but it's because I did a generate with the updated CLI, so all the examples were modified


joerdav:
Looking at the moved code, I believe a lot of this is useful for developers using templ. Not to mention these functions are already a part of our public API now, so the movement of these would likely break a fair amount of peoples code.

I think if we did go for internalising some functions we should discuss them on an individual basis, in an issue, to ensure it's definitely needed.

a-h:
@joerdav is right. You seem to have accidentally moved parts of the public API.

Rather than start with a large pull request, I think it would be best to start with a discussion at https://github.com/a-h/templ/discussions to outline what you see the issue being, and then shape a solution from there.

The `templ` package is the runtime part of templ, I'm not sure what you expect to see in there, why you think that moving it to another package is helpful, why you think it's "bloatware",  and how backwards compatibility would be maintained.

As it stands, I can't accept this PR.

otaxhu:
> Looking at the moved code, I believe a lot of this is useful for developers using templ. Not to mention these functions are already a part of our public API now, so the movement of these would likely break a fair amount of peoples code.
> 
> I think if we did go for internalising some functions we should discuss them on an individual basis, in an issue, to ensure it's definitely needed.

According to this [page (Go official documentation)](https://go.dev/doc/modules/version-numbers), the version v0 let's you do breaking changes and don't guarantee backwards compatibility. I'm still going to develop the fork, if there is some open issue related to moving functions to a internalized package I will be happy to open another pull request

otaxhu:
also you should upgrade to v1 if you want to guarantee backwards compatiblity

otaxhu:
Also sorry for my english, I used google translate, I'm so sorry if there was some misunderstanding. have a good day

PR #276, opened 2023-11-01
chore: update flake vendor hash

angaz:
Update the flake's go build vendor hash after ab819d4b1698f8f9463acb4699385372fcf89d0d.

I tested both packages build properly.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,7 +36,7 @@
           name = "templ";
           src = gitignore.lib.gitignoreSource ./.;
           subPackages = [ "cmd/templ" ];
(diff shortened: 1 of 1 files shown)


a-h:
Thanks for that. I guess we need to update the CI pipeline to test that.

angaz:
Yeah I was trying to see how it could be done. But I'm not very familiar with GitHub CI, so I just left that part out of the PR.

a-h (approved):

a-h:
The answer with Nix is always more Nix.

I've configured the build pipeline to use the Cachix Nix action and the Magix Nix Cache to speed builds up.

PR #277, opened 2023-11-02
fix: set content type correctly on gin example

veloii:
When using the gin example, the response is plain text and does not get rendered as HTML, due to the `WriteContentType` method not being called.


a-h (approved):

a-h:
Thanks for that!

Issue #279, opened 2023-11-03
Add generation warning for legacy template call expression.


joerdav:
Currently template supports two forms of template call expressions.

The original legacy format:

```
{! myComponent }
```

And the newer format which was designed to be more ergonomic with template block expressions:

```
@myComponent
@myLayout {

}
```

I don't believe that templ should support both of these, but to avoid breaking existing users I think it would make sense to do one or a combination of the following:

- An LSP warning when using the legacy format.
- An LSP code action to fix the legacy format.
- A warning on `templ generate` when using the legacu format.
- A command to fix all occurences in a project (could leverage `templ migrate`)

Along with a plan of when/if to remove it in a future major release.

a-h:
Closed in #280

PR #280, opened 2023-11-03
chore: remove reference to legacy template call expression

joerdav:
This relates to #279

### DIFF
--- a/docs/docs/03-syntax-and-usage/09-template-composition.md
+++ b/docs/docs/03-syntax-and-usage/09-template-composition.md
@@ -66,17 +66,17 @@ The use of the `{ children... }` expression in the child component.
 
 # Components as parameters
 
-Components can also be passed as parameters and rendered using the `{! component }` expression.
+Components can also be passed as parameters and rendered using the `@component` expression.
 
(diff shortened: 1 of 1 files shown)


docs/docs/03-syntax-and-usage/09-template-composition.md
```diff
@@ -66,17 +66,17 @@ The use of the `{ children... }` expression in the child component.
 
 # Components as parameters
 
-Components can also be passed as parameters and rendered using the `{! component }` expression.
+Components can also be passed as parameters and rendered using the `@component` expression.
```
a-h, line 69: Thinking about this, it probably needs some more description.

If `l` and `r` are template parameters, this syntax is correct, but if you have a `templ l() { <div>Hello</div> }`, then you need to use `@l()`, not just `@l`.

joerdav: I've not changed the description, but maybe these examples now make it clearer.

a-h (approved):

skydiver1980 (approved):

PR #284, opened 2023-11-05
refactor: remove class name sanitization, since CSS class names are already HTML attribute escaped

### LINKED ISSUE #211: Can we add an option to disabled SafeClass ?
The idea of not trusting user input by default is a good one.
Especially on suspicious character.

But for me I use a lot of TailwindCSS and the need to wrap nearly every class with `SafeClass` is overwhelm.

I don't want to change the default behavior of Templ.
No I prefer to propose to add an option to disable this behaviour so code using a lot of utility function with special character can be easy and quick to write.

And also with this option we can Imagine add an `EscapeClass` function to enable back the escape on specific part of our code.

Oh and when I say *"an option"*, I mean in the `templ generate` command.

If necessary I will obviously see how to implement this but first I would like to know if this is something that is of interest or not.

a-h:
Fixes #211


gungun974:
This PR is good. 
I think it's still good to keep some form of sensitization *(even if indirect)* for security.

The only concern here I don't see any test testing some TailwindCSS syntax. 

In Tailwind for example we use a lot of of closse with `-` and `:` like `hover:bg-sky-700`.
And even sometime thing like `text-[#50d71e]` or even `w-[calc(100%-4rem)` with the JIT.

I can suggest to be sure add some complex Class name in the generator `test-css-usage` tests.

PR #285, opened 2023-11-05
Allow script components to be rendered inline

### LINKED ISSUE #231: How to inject a standalone <script> tag with data injected
Relevant guide I was following: https://templ.guide/syntax-and-usage/script-templates#script-templates

I'm trying to add JavaScript at the end of the HTML (not inject it into an HTML attribute)

```go
package main

script example(message string) {
    console.log(message);
}

templ page(message string) {
    <html>
        <body>
            <script>
                { example(message) }
            </script>
        </body>
    </html>
}
```

I'm not sure if this is possible with Templ, but it'd be really nice if it was! Maybe I'm holding it wrong?? Any pointers would be appreciated.

### Expected

If passed in `Hello, world!`:

```html
    <html>
        <body>
            <script>
                console.log("Hello, world!");
            </script>
        </body>
    </html>
```

### Actual

If passed in `Hello, world!`:

```html
    <html>
        <body>
            <script>
                { example(message) }
            </script>
        </body>
    </html>
```

JustDerb:
Allows templ to render `script` components, which is useful for inline Javscript. As of now, `script` components can really only be used within HTML attributes. This extends the `!` usage to now enable embedding Javascript outside HTML attributes.

This fixes #231

I've also added a test case to show how it would be used (and what would be rendered). I've ensured that things should be backwards compatible, the only things that new generator runs would introduce is the new `CallInline` struct member (which is only called if you use this new functionality).


docs/docs/03-syntax-and-usage/11-script-templates.md
```diff
@@ -103,3 +103,50 @@ func main() {
...
 }
 ```
+
+You can also directly render the Javascript function, passing in Go data, using the `!` expression:
+
+```templ
+package main
+
+script printToConsole(content string) {
+	console.log(conent)
```
a-h, line 113:
```suggestion
	console.log(content)
```

docs/docs/03-syntax-and-usage/11-script-templates.md
```diff
@@ -103,3 +103,50 @@ func main() {
...
+}
+
+templ page(content string) {
+	<html>
+		<body>
+		  {! printToConsole(content) }
```
a-h, line 119: The `{! templComponent() }` syntax is old syntax that (not relevant here) doesn't support the use of child elements.

In the case of this design, there's no need for child templ elements, so maybe the new syntax of `@printToConsole(content)` would be good to use?

@joerdav has been going through and removing all the references to the old syntax. https://github.com/a-h/templ/pull/280 - but we still need to support the legacy syntax.

In terms of the implementation, it took me a couple of readthroughs to realise that you'd simply implemented the `templ.Component` interface on the `ScriptComponent`. Very smart!

runtime.go
```diff
@@ -471,13 +471,33 @@ type SafeURL string
 
 // Script handling.
 
-// SafeScript encodes unknown parameters for safety.
-func SafeScript(functionName string, params ...interface{}) string {
+func safeEncodeScriptParams(escapeHtml bool, params ...interface{}) []string {
```
a-h, line 474: This can lose the `...` and the `interface{}` since we don't need variadics here, and we're now supporting a later version of Go where we can use `any` instead.

```suggestion
func safeEncodeScriptParams(escapeHtml bool, params []any) []string {
```

runtime.go
```diff
@@ -471,13 +471,33 @@ type SafeURL string
...
 	encodedParams := make([]string, len(params))
 	for i := 0; i < len(encodedParams); i++ {
 		enc, _ := json.Marshal(params[i])
-		encodedParams[i] = EscapeString(string(enc))
+		if escapeHtml {
+			encodedParams[i] = EscapeString(string(enc))
+		} else {
+			encodedParams[i] = string(enc)
+		}
```
a-h, line 482: This could be de-nested.

```suggestion
		if !escapeHtml {
			encodedParams[i] = string(enc)
			continue
		}
		encodedParams[i] = EscapeString(string(enc))
```

runtime.go
a-h, line 474: In Go, acronyms are usually written as `escapeHTML` instead of `escapeHtml`

runtime.go
```diff
@@ -550,9 +570,33 @@ type ComponentScript struct {
 	Name string
 	// Function to render.
 	Function string
-	// Call of the function in JavaScript syntax, including parameters.
+	// Call of the function in JavaScript syntax, including parameters, and ensures parameters are HTML escaped.
 	// e.g. print({ x: 1 })
 	Call string
+	// Call of the function in JavaScript syntax, including parameters. Doesn't not HTML escape parameters; useful for directly calling.
```
a-h, line 576: "Doesn't not HTML escape parameters"?

I think it should be "Does not HTML escape parameters.".

I think it's not clear why you'd want to use `Call` vs `CallInline` from the docstring. They're both a call in JavaScript, but one is intended for calling the component script within HTML event handler attributes, and the other is intended for calling the component script from anywhere.

The docstring example I added `{ x: 1}` doesn't make it clear that params would be HTML escaped in the `Call` string at all, which was probably OK before, but now there's a `CallInline`, it might make sense to clarify.

docs/docs/03-syntax-and-usage/11-script-templates.md
```diff
@@ -103,3 +103,50 @@ func main() {
...
+		</body>
+	</html>
+}
+```
+
+The data passed into the Javascript funtion will be JSON encoded, which then can be used inside the function.
+
+```go
```
a-h, line 127: This is great, in that it shows how to use it, but it doesn't show the expected HTML output, which I think is really useful.

A lot of people aren't going to copy and paste the example and try it out, so giving the example output makes it a lot less abstract when people are skimming through.

docs/docs/03-syntax-and-usage/11-script-templates.md
```diff
@@ -103,3 +103,50 @@ func main() {
 	}
 }
 ```
+
+You can also directly render the Javascript function, passing in Go data, using the `!` expression:
```
a-h, line 107: Is it worth calling out that `ScriptComponents` are templ Components, and can therefore be included in the page?

Aside from that, it probably is worth explaining that the JavaScript function code will be written once to the page, but that the function will be called every time its parent component is rendered.

An example showing this (with output) would be very useful.

runtime.go
```diff
@@ -550,9 +570,33 @@ type ComponentScript struct {
...
+	// e.g. print({ x: 1 })
+	CallInline string
+}
+
+var _ Component = ComponentScript{}
+
+func (c ComponentScript) Render(ctx context.Context, w io.Writer) error {
```
a-h, line 583: Very clever! 👍🏻 💯

a-h (changes requested):
Brilliant. It's a better idea than what I had on how to achieve the same goal.

Well thought out and done. Just a few minor things on docs etc., and it's good to go. Thanks for your work on this.

docs/docs/03-syntax-and-usage/11-script-templates.md
JustDerb: Hmm... so swapping it to `@` in the tests seems to have broken it: 

```go
package testscriptusage

script withParameters(a string, b string, c int) {
	console.log(a, b, c);
}

script withoutParameters() {
	alert("hello");
}

templ InlineJavascript(a string) {
	{ @withoutParameters() }
	{ @withParameters(a, "test", 123) }
	// Call once more, to ensure it's defined only once
	{ @withoutParameters() }
	{ @withParameters(a, "test", 123) }
}
```

```
$ go run ./cmd/templ generate -include-version=false && go test ./...
failed to process path: ./templ/generator/test-script-inline/template.templ source formatting error: 43:36: illegal character U+0040 '@' (and 7 more errors)
```

I'll try and see what needs to be fixed to allow the newer syntax :thinking:

JustDerb: Never mind :facepalm: I realized `{ @function() }` should be simply `@function()`

docs/docs/03-syntax-and-usage/11-script-templates.md
JustDerb: Added the output

docs/docs/03-syntax-and-usage/11-script-templates.md
JustDerb: With the new output, I updated the example to call the script twice, to show it's emitted once.

runtime.go
JustDerb: Luckily I found that this approach could be used pretty quickly from some testing :smile:

runtime.go
JustDerb: The original doc string doesn't make too much sense on my first read, hence why the copy/paste. I'll update both to be clearer. I'm also not feeling super strongly on the `Call` versus `CallInline` naming and def open to better names.

runtime.go
JustDerb: I wasn't able to drop the `...` but was able to replace `interface{}`. Switching to a slice causes tons of errors across the codebase.

a-h (approved):

JustDerb:
Looks like switching to slices might have broken something... Will look on Sunday 👀

a-h:
Yes, that was a bad idea of mine then, sorry! :) thanks for looking into it!

JustDerb:
Alright, should be fixed!

a-h (approved):

a-h:
Thanks!

seanmorton:
I literally just ran into this use case today while learning templ, what a coincidence and thanks for this!

PR #288, opened 2023-11-07
Correct class value in 09-template-composition.md

andradei:
`class="right"` should be `class="contents"` based on the components defined at the beginning of the *Components as parameters* section.


a-h (approved):

a-h:
Thanks for that!

Issue #293, opened 2023-11-09
Html atributes get escaped on formatting


ross96D:
Given this code i would write using alpinejs

![image](https://github.com/a-h/templ/assets/79549577/b8314545-8768-47f9-a2fb-edbd18bddc4a)

When saving the file, so the lsp formating is call and it writes back this
![image](https://github.com/a-h/templ/assets/79549577/ea05f09a-2aa6-434d-9294-9f4dc81d8d1a)

I find this behavior uncomfortable. I would prefer the escape happens at generation time, and do no affect my templ file so i dont have to interpret what these characters mean.

I was looking at the source code and was able to craft a solution. I would be glad to submit a pull request if this behavior was not intended

OrlandoCo:
Agree, most of the time attributes are not client inputs, and maybe if required is possible to create a helper like `templ.Escape`

joerdav:
I believe this is only an issue with constant attributes `attr="somestring"`, dynamic attrs (`attr={s}`) are already ran through `templ.EscapeString`.

This format feature was introduced recently in the html format package that templ depends on: https://github.com/a-h/htmlformat/commit/5bd994fe268e4d505a9793143fa85414c7d50887

One way to avoid this in your case @ross96D would be to define your string as a go string `x-data={"{ var1='some-text' }"}`.
Although I can see why this wouldn't be ideal.

a-h:
@joerdav - the HTML formatting change isn't actually related, the HTML format lib is only used in tests to compare minified templ output with unminified expected outputs.

Formatting in templ is done by parsing a templ file into an object model, and then writing it back out.

So, when parsing a HTML element opening tag, the attribute parser is used:

https://github.com/a-h/templ/blob/fb535826f48e00c23bc8c3d0adcd78c6fc4c1bb8/parser/v2/elementparser.go#L262C15-L282

The parser attempts to grab various types of attributes from the input,.

In the case of a constant attribute it returns a `ConstantAttribute` struct, which implements the `Attribute` interface, so it can be added to list of `Attributes` on the element struct.

Constant attribute parser: https://github.com/a-h/templ/blob/fb535826f48e00c23bc8c3d0adcd78c6fc4c1bb8/parser/v2/elementparser.go#L105-L152

Element struct: https://github.com/a-h/templ/blob/fb535826f48e00c23bc8c3d0adcd78c6fc4c1bb8/parser/v2/types.go#L392-L399

[...]

PR #295, opened 2023-11-11
fix: remove escaping for ConstantAttribute (#293)

ross96D:
Remove the html.EscapeString on ConstantAttribute according to the discussion on the issue #293


a-h (approved):

a-h:
Thanks for that!

FWIW, I usually rebase PRs from main rather than merging changes from main into a PR, since it means that the commits in the branch are only the ones for the chagnge.

PR #297, opened 2023-11-13
Changing the maximum size of an element name to 128

gungun974:
So this PR is exactly the same thing of my PR #212 except I make the maximum size twice as actual.

Why change again ? Is 32 not long enough ? Well yes and no for me...

In my project I need to create some custom element who represent an administration task form controller.
For this the name should be `<admin-create-reservation-controller>` since it's represent what the custom element do. 
***Except this example is 35 longs...***

So this is why I remade the same PR as #212.

But now I was wondering ? Why is Templ limit the maximum length of an element ? Web Component don't have any restriction on their name so why ?

### DIFF
--- a/parser/v2/elementparser.go
+++ b/parser/v2/elementparser.go
@@ -313,9 +313,9 @@ var (
 			in.Seek(start)
 			return
 		}
-		if len(suffix)+1 > 64 {
+		if len(suffix)+1 > 128 {
 			ok = false
-			err = parse.Error("element names must be < 64 characters long", in.Position())
+			err = parse.Error("element names must be < 128 characters long", in.Position())
 			return
 		}
 		return prefix + suffix, true, nil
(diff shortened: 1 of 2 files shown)


a-h:
It's really just there as a backstop to identify when the parser has probably done something I wasn't expecting. I figured it would be unexpected to have such a long element name. Maybe bounce it to 128 instead, I came up with the limit based simply on the list of HTML name.

gungun974:
> It's really just there as a backstop to let users know that the document has been parsed, but has a value that they probably weren't expecting.
> 
> I figured it would be unexpected to have such a long element name, and it would likely mean a bug.
> 
> Setting it to 128 would be fine, I came up with the limit based simply on the list of HTML names.

Hum that a nice touch to prevent some big bug. 

I setup the limit up to 128 characters so this problem should never occur again for anyone.

If I made a third PR about complaining I can't write a custom element with 129 characters. 

I think you can mock me before rejecting this third PR xD

PR #299, opened 2023-11-16
Allow functions that return (string, error)

### LINKED ISSUE #287: Allow functions that return (string, error)
Similar to `html/template`, it would be amazing to be able to also return an error from functions and then make the generated component return that error.

It would make the development experience much better for me.

stephenafamo:
Fixes #287


stephenafamo:
Ready to merge. Please review @a-h

PR #303, opened 2023-11-20
Warn when using legacy calls

joerdav:


joerdav:
I've added diagnostics both for LSP and generate command. The templ fmt command also fixes this now.

a-h (approved):
Looks great!

joerdav:
I'm not sure I understand the errors from the build!

a-h:
I switched to using Nix to build so that I could check that the Nix flake would work.

Since the `go.mod` changed, the hash of deps is different, so it was just a case of updating the `vendorHash` to the latest value. If you set it to an empty string and run the build, it will tell you what the expected version is.

My understanding is that https://github.com/nix-community/gomod2nix is more efficient because it reads the Go module flake lock file, and imports each individual Go module and version into the Nix store, vs the existing implementation which grabs all the deps. Since all the deps for each app are separate, there's no efficient caching in the current version.

I think that gomod2nix will become more standard in Nixpkgs over time, but like all things Nix, it's hard to know.

PR #309, opened 2023-12-04
fix: 'generate -watch' stops sleeping between iterations after 15min

nemars:
fix: 'generate -watch' stops sleeping between iterations after 15min

`backoff.NewExponentialBackOff()` uses `DefaultMaxElapsedTime = 15 * time.Minute`

This causes 'templ generate -watch' to stop sleeping between iterations and watch for file updates constantly which in not ideal CPU utilization.

For more details, take a look at [backoff implementation](https://github.com/cenkalti/backoff/blob/v4/exponential.go).


a-h:
Thanks a lot for that!

Issue #315, opened 2023-12-09
Content-Type header, status code, and error handling


oddwhocanfly:
Hello, i found some issues with templ handler when using options.
When both status and content type options are set, content type option is ignored. 

Template:
```
templ Home(name string) {
	<div>Hello, {name}</div>
}
```

Server:
```
func main() {
	handler := templ.Handler(
		Home("John"),
		templ.WithStatus(http.StatusOK),
		templ.WithContentType("text/csv"),
	).ServeHTTP
	router := http.NewServeMux()
	router.HandleFunc("/", handler)
	http.ListenAndServe(":8000", router)
}
```

Request and response:
```
GET http://localhost:8000

HTTP/1.1 200 OK
Date: Sat, 09 Dec 2023 01:34:34 GMT
Content-Length: 22
Content-Type: text/html; charset=utf-8
```

So i've added tests for Content-Type header to templ to investigate this and my new test failed along with another existing test:
```
templ % go test       
--- FAIL: TestHandler (0.00s)
    --- FAIL: TestHandler/handlers_can_be_configured_to_return_an_alternative_status_code (0.00s)
        runtime_test.go:450: expected content-type text/html, got 
    --- FAIL: TestHandler/handlers_can_be_configured_to_return_an_alternative_status_code_and_content_type (0.00s)
        runtime_test.go:450: expected content-type text/csv, got 

```

Turns out, that when status code option is specified, this header is written first and headers from header map are ignored, because according to `ResponseWriter` `Headers()` docs, _changing the header map after a call to WriteHeader (or Write) has no effect unless the HTTP status code was of the 1xx class or the modified headers are trailers_.

To fix this, i made `ComponentHandler` write status code after the headers were set. But something seemed off and i made test `errorComponent` render some text before returning error, like `hello` test component does. More tests failed:

```
templ % go test
--- FAIL: TestHandler (0.00s)
    --- FAIL: TestHandler/handlers_that_fail_return_a_500_error (0.00s)
        runtime_test.go:450: expected status 500, got 200
        runtime_test.go:453: expected content-type text/plain; charset=utf-8, got text/html
        runtime_test.go:460:   strings.Join({
            +   "Hello",
                "templ: failed to render template\n",
              }, "")
            
    --- FAIL: TestHandler/error_handling_can_be_customised (0.00s)
        runtime_test.go:450: expected status 400, got 200
        runtime_test.go:460:   string(
            -   "custom body",
            +   "Hellocustom body",
              )
```

[...]

Issue #249, opened 2023-10-23
feature request - update the generator to warn or upgrade if the templ runtime version is not the same version as the generator executable


a-h:
In https://github.com/a-h/templ/issues/235 there was an issue that was fixed by updating the runtime.

However, updating the generator and running `templ generator` didn't let anyone know.

It could be that the generator is updated and comes to rely on a non-API-breaking change in the templ runtime that isn't present. Either way, updating the `generate` command to:

* If the `templ generate` command version < templ runtime version - warn about mismatched versions
* If the `templ generate` command version > templ runtime version - run a `go get` automatically

One point to note is that the current Nix flake, and the local snapshot install, shows `(devel)` because it's built from source and the version number isn't applied. I'd want to fix that before doing this.

joerdav:
Sounds like a useful warn to have, I wonder if there could be a way to define version mismatches that should trigger an error rather than a simple warning. "The templ version defined in your go module contains bugs x/y/z it is strongly recommended you upgrade, to surpress this warning use --someflag".


> If the templ generate command version > templ runtime version - run a go get automatically

My other thought is that this could cause some unwanted and confusing side effects in a CI pipeline. I think just breaking or warning is safer. (maybe with some color to emphasize warnings)

PentesterPriyanshu:
can you assign this issue to me i am ready to work on it

PentesterPriyanshu:
sir i have to write the script ,which starts by extracting and comparing the version numbers of the Templ runtime and the generator executable. If the versions match, it  continues with the template generation process. However, if a version mismatch is detected, the script issues a warning to the user, indicating the disparity and advising them to upgrade the Templ runtime to the latest version, using the flag for convenience.

after the version check, warning, and potential upgrade prompt, the script proceeds to generate the required template. To maintain a clean and efficient workflow, the script utilizes temporary file management, creating a temporary file for internal processing and removing it once the template generation is complete. This process helps maintain a streamlined and organized working environment for the user.

a-h:
Yesterday, I updated the CLI to auto-increment the commit number, but since the generated files now all contain an up-to-date version number, it's causing git churn.

I think that the next step is to make including the version number (and generated date/time) in the build output optional as per https://github.com/a-h/templ/issues/123#issuecomment-1777724991

The generator.go file includes some code to write out a comment which includes a version.

```go
func (g *generator) writeCodeGeneratedComment() error {
[...]

PR #316, opened 2023-12-09
Response headers, status code, and error handling

oddwhocanfly:
Proposed handler fixes and error handling improvements for #315

The general idea is to render component into a buffer before writing it to response, so that in the event of error, error handler can write its own response with correct status code and headers. The buffer is obtained with `GetBuffer()` and component should be able to use it directly as if it obtained buffer on its own.

Additionally, handler now sets headers to header map before writing response code, and does not write response code when calling error handler so that it can set its own headers and response code.

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -69,10 +69,12 @@ const componentHandlerErrorMessage = "templ: failed to render template"
 
 // ServeHTTP implements the http.Handler interface.
 func (ch ComponentHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Set header before writing status, otherwise
+	// this header won't be written to response.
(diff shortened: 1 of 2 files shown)


a-h (approved):
I just did a minor refactor, as per @matryer's talk https://youtu.be/yeetIgNeIkc?si=7BiJMIsYe3Uvghjn&t=255

a-h:
Thanks for that, much appreciated! I'm prepping for a release, just want to do some work on https://github.com/a-h/templ/issues/249, a bit more testing, and I'll push it.

oddwhocanfly:
Thank you! Yes, i did have a feeling that wasn't an idiomatic thing in go 🙂

PR #320, opened 2023-12-09
docs: fix typo in "View model" page

farant:
Hi,

I think this should be `teamInviteComponent` instead of `teamInviteComponentModel` if I understand the code correctly. Sorry for the two PRs, just wanted to change the commit message.

Thanks!


a-h (approved):
Thanks, you're right! Thanks for fixing it.

PR #323, opened 2023-12-11
fix: nix documentation

caarlos0:
- fixes a typo in CGO_ENABLED
- fixes docs on how to use templ in a flake

PS: I'm having issues with the overlay instructions too, trying to figure out what's wrong... might be something on my config though :P

### DIFF
--- a/docs/docs/02-quick-start/01-installation.md
+++ b/docs/docs/02-quick-start/01-installation.md
@@ -59,7 +59,8 @@ This flake exposes an overlay, so you can add it to your own Flake and/or NixOS 
       inherit system;
       pkgs = import nixpkgs { inherit system; };
     });
-    templ = system: inputs.templ.${system}.templ;
+    templ = system: inputs.templ.packages.${system}.templ;
+
   in {
     packages = forAllSystems ({ pkgs, system }): {
       myNewPackage = pkgs.buildGoModule {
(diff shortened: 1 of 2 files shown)


docs/docs/02-quick-start/01-installation.md
```diff
@@ -59,7 +59,8 @@ This flake exposes an overlay, so you can add it to your own Flake and/or NixOS
       inherit system;
       pkgs = import nixpkgs { inherit system; };
     });
-    templ = system: inputs.templ.${system}.templ;
+    templ = system: inputs.templ.packages.${system}.templ;
+
```
a-h, line 63:
```suggestion
```

a-h (approved):
Thanks for that!

a-h:
On the overlay, looks like it was wrong. I checked out https://github.com/DeterminateSystems/nixos-vault-service/blob/4de9bb77b05df6aa63406d99a1bf77c439414d6e/flake.nix#L50C7-L52C9 for inspiration.

I then used `nix repl` to test locally inside of the source code directory. Here, I load the local flake, then import nixpkgs with an overlay. Then I can check that `git` (which is already part of nixpkgs) is present, and also templ has been added.

```
:lf .
(import <nixpkgs> { overlays = [ outputs.overlays.default ]; }).git
(import <nixpkgs> { overlays = [ outputs.overlays.default ]; }).templ
```

PR #325, opened 2023-12-13
Feature: default to delete orphaned files, with -keepOrphanedFiles argument flag to disable feature

YiumPotato:
Added default functionality to delete orphans on `templ generate`
image
& with the -keep-orphaned-files arg flag
image
(base2_templ.go isn't deleted)

### DIFF
--- a/cmd/templ/generatecmd/main.go
+++ b/cmd/templ/generatecmd/main.go
@@ -44,8 +44,9 @@ type Arguments struct {
 	GenerateSourceMapVisualisations bool
 	IncludeVersion                  bool
 	IncludeTimestamp                bool
-	PPROFPort                       int // PPROFPort is the port to run the pprof server on.
-	KeepOrphanedFiles               bool
+	// PPROFPort is the port to run the pprof server on.
+	PPROFPort         int
+	KeepOrphanedFiles bool
 }
 
 var defaultWorkerCount = runtime.NumCPU()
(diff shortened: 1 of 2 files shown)


cmd/templ/generatecmd/main.go
```diff
@@ -125,6 +125,38 @@ func runCmd(ctx context.Context, w io.Writer, args Arguments) (err error) {
 		p = proxy.New(args.ProxyPort, target)
 	}
 
+	if !args.KeepOrphanedFiles {
```
a-h, line 128: I don't think this is the right place to do this work, because it means walking through the directory tree twice which is relatively expensive to do.

If we can combine the deletion of orphaned files with the generation, I think that would be better. Also, I'm wondering if the deletion of the orphaned file should be a warning or some other level to make it stand out.

cmd/templ/generatecmd/main.go
```diff
@@ -44,8 +44,8 @@ type Arguments struct {
 	GenerateSourceMapVisualisations bool
 	IncludeVersion                  bool
 	IncludeTimestamp                bool
-	// PPROFPort is the port to run the pprof server on.
-	PPROFPort int
+	PPROFPort                       int // PPROFPort is the port to run the pprof server on.
```
a-h, line 47: Looks like the comment has been accidentally moved? In Go, comments above the struct field are automatically included in text editor autocomplete documentation. Moving the comment to the side breaks that behaviour.

a-h (changes requested):
Thanks for your work on this, I think it would be best to move the behaviour though, unless I'm missing something?

cmd/templ/generatecmd/main.go
YiumPotato: 100% right, I was on something. Passed the argument to the processFile function (would that be the way?), checked it and used logWarning for the message.

cmd/templ/generatecmd/main.go
YiumPotato: Oops! Did not know that. I unfixed my ocd fix.

YiumPotato:
Fixed with your helpful direction. Lmk if it still needs some work!

cmd/templ/generatecmd/main.go
```diff
@@ -220,6 +220,19 @@ func processChanges(ctx context.Context, stdout io.Writer, fileNameToLastModTime
 		if info.IsDir() {
 			return nil
 		}
+		if !keepOrphanedFiles && strings.HasSuffix(fileName, "_templ.go") {
+			// lets make sure the generated file is orphaned
```
a-h, line 224:
```suggestion
			// Make sure the generated file is orphaned
```

cmd/templ/generatecmd/main.go
```diff
@@ -220,6 +220,19 @@ func processChanges(ctx context.Context, stdout io.Writer, fileNameToLastModTime
 		if info.IsDir() {
 			return nil
 		}
+		if !keepOrphanedFiles && strings.HasSuffix(fileName, "_templ.go") {
+			// Make sure the generated file is orphaned
+			// by checking if the corresponding .templ file exists
```
a-h, line 225:
```suggestion
			// by checking if the corresponding .templ file exists.
```

cmd/templ/generatecmd/main.go
```diff
@@ -220,6 +220,19 @@ func processChanges(ctx context.Context, stdout io.Writer, fileNameToLastModTime
...
+			if _, err := os.Stat(strings.TrimSuffix(fileName, "_templ.go") + ".templ"); err == nil {
+				// the .templ file exists, so we don't delete the generated file
```
a-h, line 227:
```suggestion
				// The .templ file exists, so we don't delete the generated file.
```

a-h (approved):

PR #332, opened 2023-12-15
docs: change `go run` command

### LINKED ISSUE #331: `go run *.go` doesn't work on Windows
I haven't experienced this myself (I use macOS), but apparently, [`go run *.go` doesn't work on Windows](https://stackoverflow.com/a/28082734).

According to the same Stack Overflow answer, we can use `go run .` from the `hello-world` directory to achieve the same result.

heyajulia:
`go run *.go` doesn't work on Windows, whereas `go run .` is cross-platform and has the same effect in this case.

Fixes #331

### DIFF
--- a/docs/docs/02-quick-start/02-creating-a-simple-templ-component.md
+++ b/docs/docs/02-quick-start/02-creating-a-simple-templ-component.md
@@ -73,7 +73,7 @@ func main() {
 Running the code will render the component's HTML to stdout.
 
 ```sh
-go run *.go
+go run .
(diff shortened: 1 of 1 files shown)


heyajulia:
I was worried I messed something up because I forgot to sync your repo with my fork. I synced and re-cloned my fork, and I think everything is okay now.

The force push was to change the casing of the word "change" to better align with Conventional Commits.

a-h (approved):
Thanks for that!

PR #334, opened 2023-12-17
docs: fix wiki URL for interfaces

msound:
Golang has moved the wiki from github to go.dev

See: https://github.com/golang/go/issues/61940


a-h (approved):
Thanks!

PR #336, opened 2023-12-17
[docs] Enable templ syntax highlighting in 04-core-concepts/01-components

PizzasBear:
I noticed that the last code block in [this docs page](https://templ.guide/core-concepts/components/) doesn't have syntax highlighting.

### DIFF
--- a/docs/docs/04-core-concepts/01-components.md
+++ b/docs/docs/04-core-concepts/01-components.md
@@ -63,7 +63,7 @@ templ components can be returned from methods (functions attached to types).
 
 Go code:
 
-```
+```templ
 package main
 
 type Data struct {


docs/docs/04-core-concepts/01-components.md
```diff
@@ -63,7 +63,7 @@ templ components can be returned from methods (functions attached to types).
 
 Go code:
 
-```
+```go
```
gungun974, line 66: In the documentation we have a special highlighter to highlight code with `templ` syntax in it.

Since this code block had some `templ` syntax I think it should be better if you used `templ` instead of `go`.

PizzasBear: I saw the `Go code:`, and for some reason missed the `templ` inside...

a-h:
Thanks, much appreciated. I've updated the docs to only push on a new release, so there will be a bit of a lag until you see this go live.

Since there's a few changes to the generator (support for functions that return `(string, error)`, I want to finish off work on https://github.com/a-h/templ/issues/249 before cutting a new release.

PR #337, opened 2023-12-17
Add support for rendering raw HTML, and bi-directional support for html/template

### LINKED ISSUE #175: Improve support for template/html.HTML type
I'm using the https://github.com/gorilla/csrf library to get a csrf token for forms and it provides you with a https://pkg.go.dev/html/template#HTML when you use the https://pkg.go.dev/github.com/gorilla/csrf#TemplateField func

```go
csrf.TemplateField(r)
```

This all works fine as `HTML` is just a string but right now the compiler does give a warning or something? 

Screenshot 2023-09-29 at 20 58 33

A really simple reproducible example is 

```go
package template

templ SomePage(h template.HTML) {
    <div>
        { h }
    </div>
}
```

Which generates the following 

```go
package template

//lint:file-ignore SA4006 This context is only used if a nested component is present.

import "github.com/a-h/templ"
import "context"
import "io"
import "bytes"

func SomePage(h template.HTML) templ.Component {
	return templ.ComponentFunc(func(ctx context.Context, w io.Writer) (err error) {
		templBuffer, templIsBuffer := w.(*bytes.Buffer)
		if !templIsBuffer {
			templBuffer = templ.GetBuffer()
			defer templ.ReleaseBuffer(templBuffer)
		}
		ctx = templ.InitializeContext(ctx)
		var_1 := templ.GetChildren(ctx)
		if var_1 == nil {
			var_1 = templ.NopComponent
		}
		ctx = templ.ClearChildren(ctx)
		_, err = templBuffer.WriteString("<div>")
		if err != nil {
			return err
		}
		var var_2 string = h
		_, err = templBuffer.WriteString(templ.EscapeString(var_2))
		if err != nil {
			return err
		}
		_, err = templBuffer.WriteString("</div>")
		if err != nil {
			return err
		}
		if !templIsBuffer {
			_, err = templBuffer.WriteTo(w)
		}
		return err
	})
}
```

Templ Version 

```
❯ templ --version
v0.2.335-0.20230929162951-efd33120d9e8
```

I would try do this but I'm still getting to know the library and how it all works so I thought I'd just log it here in case anyone else with more experience can quickly solve this

a-h:
Fixes #175


docs/docs/03-syntax-and-usage/15-rendering-raw-html.md
```diff
@@ -0,0 +1,31 @@
+# Rendering raw HTML
+
+To render HTML that has come from a trusted source, bypassing all HTML escaping and security mechanisms that templ includes, use use the `templ.Raw` function.
```
muradbu, line 3: > use use the `templ.Raw` function.

You wrote "use" twice

PR #338, opened 2023-12-17
feat: show templ CLI and go.mod version mismatch warning

### LINKED ISSUE #249: feature request - update the generator to warn or upgrade if the templ runtime version is not the same version as the generator executable

a-h:
Fixes #249

Screenshot 2023-12-17 at 20 24 02

### DIFF
--- a/cmd/templ/generatecmd/main.go
+++ b/cmd/templ/generatecmd/main.go
@@ -449,6 +449,9 @@ func checkTemplVersion(dir string) error {
 				if cmp < 0 {
 					return fmt.Errorf("generator %v is newer than templ version %v found in go.mod file, consider running `go get github.com/a-h/templ`", templ.Version(), r.Mod.Version)
 				}
+				if cmp > 0 {
+					return fmt.Errorf("generator %v is older than templ version %v found in go.mod file, consider upgrading templ CLI", templ.Version(), r.Mod.Version)
+				}
 				return nil
 			}
 		}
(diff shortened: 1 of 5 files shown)


kron4eg:
After recent update of Go and running `go mod tidy` my go.mod looks like this:

```diff
diff --git go.mod go.mod
index d33db29..76bfbda 100644
--- go.mod
+++ go.mod
@@ -2,8 +2,10 @@ module exp2

 go 1.21.4

+toolchain go1.21.5
+
 require (
-       github.com/a-h/templ v0.2.476
+       github.com/a-h/templ v0.2.501
```

Which results to the error:
```
(✓) Generated code for 2 templates with 0 errors in 2.108317ms
(!) templ version check failed: failed to parse go.mod file: <...>/src/backend/go.mod:3: invalid go version '1.21.4': must match format 1.23
<...>/src/backend/go.mod:5: unknown directive: toolchain
```

a-h:
Ah yes, just noticed it. Sorry about that... #355

I think the temp solution while the go module parser issue is open is to regexp out the final part of the version before parsing.

I might have time to do that tomorrow and put out a tiny release.

PR #350, opened 2023-12-21
[docs] - Added Note for PATH variable on Readme

georgeiliadis91:


a-h:
Thanks for this.

From reading this PR, I realised that the `install-snapshot` script is putting the file in the wrong place. It works on my system, but not for everyone else.

I thought it would be better to `go install` the package instead, then the command wouldn't need a note, because it's much more likely that anyone that has Go installed has $GOBIN or $HOME/go/bin already in their path as per this page: https://pkg.go.dev/cmd/go#hdr-Compile_and_install_packages_and_dependencies

> Executables are installed in the directory named by the GOBIN environment variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH environment variable is not set. Executables in $GOROOT are installed in $GOROOT/bin or $GOTOOLDIR instead of $GOBIN.

georgeiliadis91:
cheers, appreciate that you included it in the commit chain.

a-h:
Thanks for the contribution, and for taking the time to say that! I think it's the right thing to do. You put work in, I make sure you get listed as a contributor.

I've been on the other side, where I've put work into something and then a maintainer folds it in without credit, which I think is a very uncool thing to do. 😀

PR #351, opened 2023-12-21
Remove extra line between comments and templ code

### LINKED ISSUE #335: Comments inside `.templ` files above templ component get an extra line
I've search the docs and issues, apparently (maybe I've missed it) no one has reported it. Is this behaviour intentional due to some reasons?

Thank you!

templ-comment-fmt

edmondwinston:
## What's changed?

This PR adds conditions to check if above a Templ block exists comments before appending new line characters.
Fixes #335.

## POC

- When running `fmt`:

templ-comment-fmt

- When running `fmt`:

templ-comment-generate


edmondwinston:
Your suggestion made it much easier, thank you.

edmondwinston:
Now looking at the generator's testing part, what would be the alternative for making a new directory, create a `.templ` file, render it, parse the generated file, then compare the content? I understand the answer is somewhere in the codebase, but do you happen to have it on the top your head?

a-h:
Thanks for the PR!

On the generator testing, I'm not sure I follow what you're looking to test. There's already a test in the CI that ensures that all _templ.go files are up-to-date.

edmondwinston:
@a-h

I meant to add test cases to make sure that the generated comments would also not have that extra line break, for hover docs purpose. It's fine for now, but just to make sure it won't, I think.

a-h:
I'm very sorry, but I'm still not understanding the test scenario. If you're going to follow up with some extra tests, brilliant, I'm sure that will explain it more precisely than English can. 😁

Issue #349, opened 2023-12-20
Go Fiber Support / Example?


jacobranson:
Corresponding ticket on Go Fiber's side: https://github.com/gofiber/template/issues/302

Based on some of the conversation going on over there, it looks like there are a few different approaches to integrating both Templ and Fiber. Since there are already examples of Gin and Chi, maybe one could be added for Fiber?

a-h:
I think that since the Fiber context type implements io.Writer, it should be a case of:

```go
app.Get("/", func(c *fiber.Ctx) error {
  return templConponent("hello").Render(c.Context(), c)
})
```

But I'm just typing this on a mobile, haven't had chance to try it!

bastianwegge:
I created a helper that does this:

```go
func Render(c *fiber.Ctx, component templ.Component, options ...func(*templ.ComponentHandler)) error {
	componentHandler := templ.Handler(layout.Main(component))
	for _, o := range options {
		o(componentHandler)
	}
	return adaptor.HTTPHandler(componentHandler)(c)
}
```

This way you can use your render-function like this, which I found easiest:

```go
app.Get("/", func(c *fiber.Ctx) error {
  return view.Render(c, MyTemplView())
})
```

a-h:
I think we can consider this closed. We've got a good example now, thanks to @bastianwegge.

PR #356, opened 2023-12-22
docs: add simple example for integration with gofiber

bastianwegge:
As requested and discussed in these Issues I created a small example for an integration of Templ into gofiber. The example currently goes without a README.md file, I oriented on the other examples you currently provide.

Issues:
- https://github.com/a-h/templ/issues/349
- https://github.com/gofiber/template/issues/302#issuecomment-1867958802

If there's any more you'd like me to do before this goes in, please go ahead. I could also provide other examples regarding CSRF and HTMX or a small Templ gofiber HTMX Todo app.


a-h:
Thanks for that, much appreciated!

PR #357, opened 2023-12-23
docs: add star history for templ

knudtty:
Templ is growing quickly and getting a lot of love, let's add the star history to show that!


a-h:
Hi, thanks for the PR, but I don't see the point of putting the star history graph in the README. The star count is already visible in Github, so let's save people the bandwidth of downloading the SVG. 😁

Issue #318, opened 2023-12-09
Proposal: DEV Mode


stephenafamo:
Right now, development is sort of slow with `templ` because each change requires my entire application to be recompiled. For example, changing a class on a `div` requires 5 seconds before I can see it in my browser.

I would suggest having a "DEV mode" e.g. `templ generate --watch --dev-mode`

In this mode, instead of fully generating the static template code, some parts of the code are generated in the filesystem. This way, a full rebuild is not required each time something in the template changes.

I understand that this is difficult in reality since the templ code interfaces with the Go code, but I think there is a way to make this work.

1. The template is split into "static strings" and "go code".
2. Anytime part of the "go code" changes, then a full rebuild is required.
3. The static strings are put in a separate file where they are read from each time. This can be modified without needing the application to be rebuilt.

If it will be too complex to maintain both rendering modes, they could be combined such that in **dev mode**, the filesystem is read each time, but in **prod mode** the static strings are embedded with `//go:embed`

a-h:
Great idea, and great ideas on the implementation too - it sounds quite practical.

I think that two generation modes would work best, since otherwise we'd have to have the concept of a "StringProvider" - something to look up the values of strings by their integer IDs so that we could swap implementation between dev and prod. Having a `StringProvider` would harm runtime performance, since we'd go from pushing a string constant, to making a virtual function call, but having two generation modes isn't too painful.

The change would be in this part: https://github.com/a-h/templ/blob/ddb1610a81627f5c9c524bb66877477f401e4ecd/generator/generator.go#L1280-L1294

We'd have to add a mode to the generator struct at https://github.com/a-h/templ/blob/ddb1610a81627f5c9c524bb66877477f401e4ecd/generator/generator.go#L55 and possibly add a `Strings []string` field to it. When we generate the file in dev mode, instead of loading the string directly, the generator would have to write out some code that requests a string by its id - for example, `templ_var_1, err := templdevmode.GetString("./{relative_path}/name.templ", templ_templateHash421, 1)`.

Since the generator adds strings to the `Strings` slice, in dev mode, the `templ generate` command can be updated to write them to disk (maybe as `{template_name}.templdev` so they can be added to `.gitignore`/`.ignore` files and/or deleted), and the `GetString` function can load them from disk on every render.

[...]

PR #366, opened 2023-12-27
Improve watch mode to reduce recompilation

stephenafamo:
Implementation for #318 and #340


stephenafamo:
Not 100% ready. Needs some cleaning up and testing.

Also, I can't figure out why, but additional builds in watch mode seem to ignore the string extraction.

stephenafamo:
> additional builds in watch mode seem to ignore the string extraction.

It seems some "phantom `templ generate`" is also running in the background and it overwrites the file.   
I am not sure what invokes it, but it seems to use some existing `templ` because it ignores things like adding a custom comment at the end of each line.

stephenafamo:
> but it seems to use some existing `templ`

This was the issue. I had a running `templ generate --watch` that wasn't killed properly. The PR works great 👌🏾

stephenafamo:
The PR is now ready to be reviewed @a-h 

Side note: I also modified `writeText` to use `WriteLiteralString`

stephenafamo:
Waiting on a review 🙏🏾 @a-h

a-h:
Apologies. I want to have a good sit down and really understand it. With the holidays, I've just been picking up small stuff.

I will get on it though, I'm excited about the change. Think it will be a big improvement to dev performance.

cmd/templ/generatecmd/main.go
```diff
@@ -129,15 +147,36 @@ func runCmd(ctx context.Context, w io.Writer, args Arguments) (err error) {
...
+	start := time.Now()
+
 	bo := backoff.NewExponentialBackOff()
 	bo.InitialInterval = time.Millisecond * 500
 	bo.MaxInterval = time.Second * 3
 	bo.MaxElapsedTime = 0
 
 	var firstRunComplete bool
 	fileNameToLastModTime := make(map[string]time.Time)
+	fileNameToHash := make(map[string][md5.Size]byte)
```
a-h, line 172: We're not using MD5 for security here, but use of it will flag as a deprecated / insecure hashing algorithm in security reviews, so I'd prefer to use SHA256.

I doubt there's much difference in the performance of MD5 vs SHA256, given that Go uses specific processor instructions for architectures that support it.

https://cs.opensource.google/go/go/+/master:src/crypto/sha256/sha256block_amd64.go

https://cs.opensource.google/go/go/+/master:src/crypto/sha256/sha256block_arm64.go

cmd/templ/generatecmd/main.go
```diff
@@ -67,29 +70,44 @@ func Run(w io.Writer, args Arguments) (err error) {
...
-			fmt.Fprintln(w, "\nCancelling...")
-			err = run.Stop()
-			if err != nil {
-				fmt.Fprintf(w, "Error killing command: %v\n", err)
+		watching := args.Watch
+		for {
+			select {
+			case <-signalChan: // First signal, cancel context.
+				if watching {
+					fmt.Println("stop watching")
```
a-h, line 80:
```suggestion
					fmt.Println("Stopping watch operation...")
```

cmd/templ/generatecmd/main.go
```diff
@@ -179,20 +218,62 @@ func runCmd(ctx context.Context, w io.Writer, args Arguments) (err error) {
 				}()
 			}
 		}
-		if err = checkTemplVersion(args.Path); err != nil {
+		if err := checkTemplVersion(args.Path); err != nil {
```
a-h, line 221: Nice!

cmd/templ/generatecmd/main.go
```diff
@@ -179,20 +218,62 @@ func runCmd(ctx context.Context, w io.Writer, args Arguments) (err error) {
...
+
 		firstRunComplete = true
 		start = time.Now()
 	}
-	return err
+
+	return nil
+}
+
+func generateProduction(ctx context.Context, w io.Writer, args Arguments, opts []generator.GenerateOpt, p *proxy.Handler) error {
```
a-h, line 239: I like how this is separate. It keeps the risk of breaking stuff in dev mode relatively low.

cmd/templ/generatecmd/main.go
```diff
@@ -179,20 +218,62 @@ func runCmd(ctx context.Context, w io.Writer, args Arguments) (err error) {
...
+		if args.Command != "" {
+			fmt.Fprintf(w, "Executing command: %s\n", args.Command)
+			if _, err := run.Run(ctx, args.Path, args.Command); err != nil {
+				fmt.Fprintf(w, "Error starting command: %v\n", err)
+			}
+		}
+		// Send server-sent event.
+		if p != nil {
+			p.SendSSE("message", "reload")
+		}
```
a-h, line 269: This shouldn't be required in prod mode?

cmd/templ/generatecmd/main.go
```diff
@@ -210,10 +291,18 @@ func shouldSkipDir(dir string) bool {
 	return false
 }
 
-func processChanges(ctx context.Context, stdout io.Writer, fileNameToLastModTime map[string]time.Time, path string, generateSourceMapVisualisations bool, opts []generator.GenerateOpt, maxWorkerCount int, keepOrphanedFiles bool) (changesFound int, errs []error) {
+func processChanges(ctx context.Context, stdout io.Writer, fileNameToLastModTime map[string]time.Time, hashes map[string][md5.Size]byte, path string, generateSourceMapVisualisations bool, opts []generator.GenerateOpt, maxWorkerCount int, watching, keepOrphanedFiles bool) (changesFound int, errs []error) {
```
a-h, line 294: This is starting to look a mess with so many params being thrown around, but I think it's best to deal with that in a separate refactor.

cmd/templ/generatecmd/main.go
```diff
@@ -348,8 +447,26 @@ func generate(ctx context.Context, basePath, fileName string, generateSourceMapV
 		return nil, fmt.Errorf("%s source formatting error: %w", fileName, err)
 	}
 
-	if err = os.WriteFile(targetFileName, data, 0644); err != nil {
-		return nil, fmt.Errorf("%s write file error: %w", targetFileName, err)
+	// Hash and write the file if the hash has changed
```
a-h, line 450:
```suggestion
	// Hash output, and write out the file if the hash has changed.
```

cmd/templ/generatecmd/main.go
```diff
@@ -348,8 +447,26 @@ func generate(ctx context.Context, basePath, fileName string, generateSourceMapV
...
+	hash := md5.Sum(data)
```
a-h, line 451: `data` is not a good variable name. Not an issue in this PR, but I'd prefer it the name was `formattedGoCode`.

cmd/templ/generatecmd/main.go
```diff
@@ -348,8 +447,26 @@ func generate(ctx context.Context, basePath, fileName string, generateSourceMapV
...
+	if hashes[targetFileName] != hash {
+		if err = os.WriteFile(targetFileName, data, 0o644); err != nil {
+			return nil, fmt.Errorf("%s write file error: %w", targetFileName, err)
+		}
+		hashes[targetFileName] = hash
+	}
+
+	// Add the txt file if it has changed
```
a-h, line 459:
```suggestion
	// Add the txt file if it has changed.
```

cmd/templ/generatecmd/main.go
```diff
@@ -348,8 +447,26 @@ func generate(ctx context.Context, basePath, fileName string, generateSourceMapV
 		return nil, fmt.Errorf("%s source formatting error: %w", fileName, err)
 	}
 
-	if err = os.WriteFile(targetFileName, data, 0644); err != nil {
-		return nil, fmt.Errorf("%s write file error: %w", targetFileName, err)
+	// Hash and write the file if the hash has changed
+	hash := md5.Sum(data)
+	if hashes[targetFileName] != hash {
+		if err = os.WriteFile(targetFileName, data, 0o644); err != nil {
+			return nil, fmt.Errorf("%s write file error: %w", targetFileName, err)
```
a-h, line 454: All the other errors in templ are basically of the form `failed to ...`, so I'd prefer:

```suggestion
			return nil, fmt.Errorf("failed to write target file %q: %w", targetFileName, err)
```

cmd/templ/generatecmd/main.go
```diff
@@ -348,8 +447,26 @@ func generate(ctx context.Context, basePath, fileName string, generateSourceMapV
...
+	}
+
+	// Add the txt file if it has changed
+	if len(literals) > 0 {
+		txtFileName := strings.TrimSuffix(fileName, ".templ") + "_templ.txt"
+		contents := strings.Join(literals, "\n")
+		txtHash := md5.Sum([]byte(contents))
+		if hashes[txtFileName] != txtHash {
+			if err = os.WriteFile(txtFileName, []byte(strings.Join(literals, "\n")), 0o644); err != nil {
+				return nil, fmt.Errorf("%s write static text error: %w", txtFileName, err)
```
a-h, line 466:
```suggestion
				return nil, fmt.Errorf("failed to write string literal file %q: %w", txtFileName, err)
```

generator/generator.go
```diff
@@ -47,9 +47,17 @@ func WithFileName(name string) GenerateOpt {
 	}
 }
 
+func WithExtractStrings() GenerateOpt {
+	return func(g *generator) error {
+		g.w.extractStrings = true
+		g.w.strings = []string{""}
```
a-h, line 53: This looks weird to me. Does it really need to have an empty string in it? No big deal, just a head scratcher. 😁

generator/generator.go
```diff
@@ -62,6 +70,8 @@ func Generate(template parser.TemplateFile, w io.Writer, opts ...GenerateOpt) (s
 	}
 	err = g.generate()
 	sm = g.sourceMap
+	literals = g.w.strings
+
```
a-h, line 74:
```suggestion
```

generator/generator.go
```diff
@@ -1350,19 +1360,9 @@ func (g *generator) writeWhitespace(indentLevel int, n parser.Whitespace) (err e
...
-	}
-	// _, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString(vn)
-	if _, err = g.w.WriteIndent(indentLevel, "_, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString("+vn+")\n"); err != nil {
-		return err
-	}
-	if err = g.writeErrorHandler(indentLevel); err != nil {
-		return err
-	}
-	return nil
+	quoted := strconv.Quote(n.Value)
```
a-h, line 1363: 🥇 Nicely done. Makes me wonder whether `createGoString` should die... Can be done in a separate refactor though.

generator/rangewriter.go
```diff
@@ -17,13 +18,21 @@ type RangeWriter struct {
 	Current   parser.Position
 	inLiteral bool
 	w         io.Writer
+
+	// Extract strings
```
a-h, line 22:
```suggestion
	// Extract strings.
```

benchmarks/templ/template_templ.go
a-h, line 1: Looks like an improvement to the overall string literal handling, which should give a little boost to performance, and a reduction in size of generated Go code. Nice.

runtime.go
```diff
@@ -777,3 +781,36 @@ func ToGoHTML(ctx context.Context, c Component) (s template.HTML, err error) {
 	s = template.HTML(b.String())
 	return
 }
+
+// WriteExtractedString is used when rendering templates in development mode.
+// the generator would have written non-go code to the _templ.txt file, which
+// is then read by this function and written to the output.
+func WriteExtractedString(w *bytes.Buffer, lineNum int) error {
```
a-h, line 788: I understand why the name is this, but in my mind, it makes sense to be called `GetWatchModeString` or something.

I think that this should be optimised further to maintain a global variable for last modified date on the strings file, and to load the whole file into RAM, so we don't hit the disk as often, since with this design it looks like each string literal would result in a filesystem read.

So I think... split the file into lines, and keep the array of strings. Reload the array of strings if the file has changed since we last read it.

generator/rangewriter.go
```diff
@@ -17,13 +18,21 @@ type RangeWriter struct {
...
+	extractStrings bool
+	strings        []string
 }
 
 func (rw *RangeWriter) closeLiteral(indent int) (r parser.Range, err error) {
 	rw.inLiteral = false
-	_, err = rw.write("\")\n")
-	if err != nil {
-		return
+	if rw.extractStrings {
```
a-h, line 29: I don't understand what's going on here. Probably because I haven't read the rest yet, but maybe a comment would help.

generator/rangewriter.go
```diff
@@ -48,15 +57,30 @@ func (rw *RangeWriter) WriteStringLiteral(level int, s string) (r parser.Range,
...
+				return
+			}
 		}
 	}
-	_, err = rw.write(s)
-	if err != nil {
-		return
+
+	if rw.extractStrings {
+		rw.strings[len(rw.strings)-1] += s
```
a-h, line 74: Ah, now I see why the strings slice is initialised with an empty string. I wonder now whether having a slice of strings.Builder would be more efficient in terms of memory allocations.

I think it might be better.

As you likely know, concatenating strings causes a new string to be created that's the size of both strings.

In small strings, the cost of initialising the string builder can be higher.

generator/rangewriter.go
```diff
@@ -48,15 +57,30 @@ func (rw *RangeWriter) WriteStringLiteral(level int, s string) (r parser.Range,
...
 		}
-		if _, err = rw.WriteIndent(level, `_, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString("`); err != nil {
-			return
+
+		if rw.extractStrings {
+			index := len(rw.strings)
+			if _, err = rw.WriteIndent(level, "templ_7745c5c3_Err = templ.WriteExtractedString(templ_7745c5c3_Buffer, "+strconv.Itoa(index)+")\n"); err != nil {
+				return
+			}
+		} else {
```
a-h, line 66: When I see if/else to enable/disable a feature, I usually wonder whether I'm missing an abstraction. Should there be a LiteralWriter interface that has two implementations (Production, and WatchMode), or has my mind been warped by too much C#. 😁 

I'm happy with whatever you think best here.

a-h (changes requested):
Wow, really awesome work, thanks so much for putting this together, and I'm really sorry it took me so long to get to looking at this.

I think there's a few bits to clear up, but it's brilliant.

I've asked for a few changes, and put in a few suggestions. Once that's sorted, I'll merge it, and put together an integration test as per https://github.com/a-h/templ/issues/381#issuecomment-1877086362

Again, thanks so much.

generator/generator.go
stephenafamo: Looks like it could be part of the string literals. I will take a look at this in a different PR.

runtime.go
stephenafamo: I renamed it, but to `WriteWatchModeString` instead since it doesn't return the string but writes it.

I will think of a way to maintain the state so that we don't hit the filesystem as often. We would need to maintain a struct for each file and somehow initialize it at the start.

generator/rangewriter.go
stephenafamo: Took your suggestion and used an interface with implementations. Works nicely 👌🏾

generator/rangewriter.go
stephenafamo: Used a `strings.Builder` and got rid of the slice and string concatenation.

stephenafamo: But I use only one builder and just add a new line for a new literal so there is no extra cost

generator/rangewriter.go
stephenafamo: Cleaned this up

generator/generator.go
stephenafamo: Cleaned this up

runtime.go
stephenafamo: Found a way to do the caching using a global map.

I decided to leave out having a mutex since I don't think there is much danger here. The worst case is that multiple calls overwrite the same entry in the map if they're fast enough, but I don't think that's such a big problem.

stephenafamo:
Ready for another review @a-h

runtime.go
a-h: I was thinking of something like this...

```go
package templ

var watchState map[string]*WatchState
var watchStateMutex sync.Mutex = new(sync.Mutex)

type WatchState struct {
  LastMod time.Time
  Strings []string
}

func WriteWatchModeString(fileName string, index int, w io.Writer) error {
  watchStateMutex.Lock()
  defer watchStateMutex.Unlock()

  state, ok := watchState[fileName]
  if !ok {
    //TODO: Load the state and store it in the map.
    state = new(WatchState)
    //TODO: Populate the state by reading from disk.
    watchState[fileName] = state
  }
  //TODO: Re-os.Stat the file if the last mod is > 100ms in the past or something (basically, don't thrash the disk).
  //TODO: Re-read the file if it has been modified and update the state.
  //TODO: Return the string. 
} 
```

a-h: Ah, I see you beat me to it! I'll read through the updates tomorrow!

stephenafamo: Cleaned it up a bit more, introduced the 100ms buffer and a mutex

a-h:
Brilliant work, thanks a lot. I think it's worth a blog post to explain the improvement, and how it's a potential game changer for development experience with templ. If you want to write it, go ahead, if not, I will - just let me know.

The next thing to deal with is to switch out the very basic "walk the directory every second" with a filesystem watcher implementation. I couldn't get fsnotify to work properly on my Mac, but maybe I was using it wrong.

philippefuentes:
> Brilliant work, thanks a lot. I think it's worth a blog post to explain the improvement, and how it's a potential game changer for development experience with templ. If you want to write it, go ahead, if not, I will - just let me know.
> 
> The next thing to deal with is to switch out the very basic "walk the directory every second" with a filesystem watcher implementation. I couldn't get fsnotify to work properly on my Mac, but maybe I was using it wrong.
> 
> With that in place, the reload time will be incredible...

Hi @a-h what was the problems you encountered with `fsnotify` ?

stephenafamo:
> Brilliant work, thanks a lot. I think it's worth a blog post to explain the improvement, and how it's a potential game changer for development experience with templ. If you want to write it, go ahead, if not, I will - just let me know.

I will write something about this. It's been too long since I wrote a blog post and this definitely deserves one.

a-h:
@philippefuentes the issue I had was inconsistent notifications. Sometimes it would work, sometimes not.

a-h:
Basically, the watch implementation needs some automated integration tests, including one that can be ran for extended periods to identify issues that crop up over time.

PR #368, opened 2023-12-27
add importing by adding cdn link to head example

szymon676:


a-h:
I think this might be in the wrong place, since this documentation section file is about IDE support.

Also, there's a few typos in this example and a missing import, meaning it wouldn't compile.

szymon676:
oh yeah, you are right. maybe move this to another page with cdns?

PR #386, opened 2024-01-02
Update 03-hot-reload.md: Added `wgo`

guettli:
Added `wgo` as alternative to `air`.

It needs no config file, and has the MIT license.


a-h:
Brilliant. Thanks a lot!

PR #392, opened 2024-01-03
docs: add emacs to editors with templ support

danderson:
Thanks for templ! I made an emacs major mode for it, reusing the tree-sitter parser that the neovim support uses.


a-h (approved):

a-h:
Excellent! Thanks for doing that. 😁

a-h:
The changes will hit the docs at the next release of templ.

PR #400, opened 2024-01-06
docs(installation): add nixpkgs installation instruction

rsHalford:
Nix now has support for a more straightforward templ installation on the [unstable channel](https://search.nixos.org/packages?channel=unstable&show=templ&from=0&size=50&sort=relevance&type=packages&query=templ).

I've opted for the `nix-env` command, as Nix users are likely comfortable adapting it to suit their systems.

If required, I could also update this PR to present `nix-shell` and NixOS configuration installation methods. But, would like to know if the previous Nix installation methods are to be kept first.

### DIFF
--- a/docs/docs/02-quick-start/01-installation.md
+++ b/docs/docs/02-quick-start/01-installation.md
@@ -14,6 +14,12 @@ Download the latest release from https://github.com/a-h/templ/releases/latest
 
 ## Nix
 
+templ is available to install using the nixpkgs-unstable channel.
+
+```sh
+nix-env -f channel:nixpkgs-unstable -iA templ
+```
+
 templ provides a Nix flake with an exported package containing the binary at https://github.com/a-h/templ/blob/main/flake.nix
 
 ```sh


a-h:
Thanks for the PR, and for updating nixpkgs (on that, I think there's a way to automate the update, but I'm not sure how it's done).

I'm not sure about having `nix-env` as the first option. My understanding is that it installs into your current environment, much like `brew` or similar, versus maintaining a list of programs that you use in source control as per home-manager.

Perhaps leading with `nix shell` to run a shell with templ in it would be the best approach? It can use `nix shell nixpkgs#templ`, and then also explain how to run a specific commit of templ using the Flake reference, e.g. `nix shell github:a-h/templ/d6a0945906af6b31e3554d6ed8781ad83172a30c`. 

I'm usually passing people to https://zero-to-nix.com/ as a good way to learn Nix, and it uses the new CLI syntax everywhere, and uses the `nix shell` command to run programs.

Personally, I use a Flake for anything that's fast moving so I can build specific revisions from source if I need to, since sometimes PRs for updates to nixpkgs take a while to get approved, and I usually nixpkgs to a specific commit, meaning I don't get updates.

I appreciate that the docs aren't supposed to be explaining Nix to people, but anything we can do to reduce the wall of complexity around Nix is good. 😀

a-h:
This seems to have ran its course for now.

PR #427, opened 2024-01-16
feat: added format on save functionality to the lsp as requested by @anthdm

danecwalker:
Added the functionality of formating .templ files on save via the lsp as requested by @anthdm [here](https://www.youtube.com/live/i1Vf47HfnPs?si=ZKGlD_UtA-tavoPZ&t=425)

### DIFF
--- a/cmd/templ/lspcmd/proxy/format.go
+++ b/cmd/templ/lspcmd/proxy/format.go
@@ -1,0 +1,34 @@
+package proxy
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+
+	"github.com/a-h/templ/parser/v2"
+	"github.com/natefinch/atomic"
+)
+
+func format(fileName string) (err error) {
+	contents, err := os.ReadFile(fileName)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


joerdav:
I don't believe that this should be implemented here. Templ does support lsp formatting already:

https://github.com/a-h/templ/blob/main/cmd/templ/lspcmd/proxy/server.go#L720

It's the clients responsibility to request a format. VSCode has a "Format on save" function, which does in-fact work with templ.

The issue @anthdm is experiencing here is that templ isn't formatting multiline go structs as expected (or at all) I do believe this should be addressed, I'll raise an issue with a minimal reproduction I've created.

a-h:
Thanks for your contribution, but I also agree with @joerdav

Did we fix the formatting issue already @joerdav? I don't recognise the issue.

joerdav:
Yes, this was the issue of templelements note being passed through the go format function. So is fixed now. Though I think @antdm is having some different formatting issues now, just trying to get minimal reproductions that we can put into unit tests.

PR #428, opened 2024-01-16
docs: Add Emmet config for vscode

jonerer:
Just a short snippet on how to activate Emmet for templ files


a-h:
Thanks for that!

Issue #418, opened 2024-01-11
[BUG]Endless loop causes 100% CPU load and hangs templ


ngortheone:
When running `temp generate` the command hangs after generating files.
It also causes 100% CPU usage. The only way to end this is double `ctrl-c`.

I did some digging and the hang happens due to an endless for loop here.

https://github.com/a-h/templ/blob/914bc73f8030166c9f8bd46af62f5f9bb2c3b089/cmd/templ/generatecmd/main.go#L540

The code attempts to patch my `go.mod` file (something that shouldn't be default behavior and should be an explicit opt-in) and fails to find teml - because I don't have it there and the loop repeats.

benjiro29:
The watch issue is different then your problem. Its was timeout that expired after 15 minutes. And that has been solved in the latest version.

>  should be an explicit opt-in) and fails to find teml - because I don't have it there and the loop repeats.

It does not patch your actual file .. Check the code. Templ needs to read the go.mod to check the version but the go parser has a issue with xx.xx.xxx formatting. So what the program does is first patch the number, and then feed the modified go.mod file to the parser.

So no issue there...

ngortheone:
Agreed. My primary problem is a permanent hang that happens when go.mod has no `a-h/templ`  in it. I think in this case it should exit with a warning, that says something like "no templ found in your go.mod - you will be unable to compile your generated files"

a-h:
The draft PR is looking in reasonable shape now. It would be great if you could review it, or test it out.

In the PR, I've added some integration tests that run the actual generate command against a test project, including making some changes, and I've implemented something similar to @ngortheone's suggestion. In the case you don't have templ in your go.mod it will warn:

```
(!) templ version check failed: templ not found in go.mod file, run `go get github.com/a-h/templ to install it`
```

PR #441, opened 2024-01-19
fix: infinite loop if go.mod not found, warn on missing go.mod or forgotten go get

a-h:
Fixes https://github.com/a-h/templ/issues/418


joerdav:
I'm wondering if the lack of a go.mod should not be an error. Since they are an optional feature of go.

a-h:
Although it prints an error if the go.mod is not present, it doesn't actually stop anything @joerdav

joerdav:
Ah yeah, it doesn't exist just a warning log, I forgot about that one! Cool LGTM

joerdav:
One thing, the test project could go in a `testdata` folder, then it'll only ever be compiled on test runs.

PR #459, opened 2024-01-25
docs: Fixes typo on 02-blog-example.md

samueldaviddelacruz:
Fixes typo, using ! instead of @ to render body component


joerdav:
Good spot, thanks!

PR #461, opened 2024-01-25
refactor: migrate to Go parser instead of using heuristics

a-h:
This should make parsing Go code within templ files more robust, since it uses the actual Go parser.


parser/v2/conditionalattributeparser.go
```diff
@@ -12,23 +13,27 @@ func (_ conditionalAttributeParser) Parse(pi *parse.Input) (r ConditionalAttribu
...
+	if r.Expression, err = parseGo("if attribute", pi, goexpression.If); err != nil {
+		return r, false, err
 	}
 
 	// Eat " {\n".
-	if _, ok, err = until.Parse(pi); err != nil || !ok {
-		err = parse.Error("attribute if: unterminated (missing closing '{\n')", pi.Position())
+	if _, ok, err = openBraceWithOptionalPadding.Parse(pi); err != nil || !ok {
+		err = parse.Error("attribute if: unterminated (missing closing '{\n')", pi.PositionAt(start))
 		return
```
joerdav, line 32: This function now has a mix of naked/explicit returns. (potentially a nit pick)

parser/v2/goexpression/parse.go
```diff
@@ -0,0 +1,257 @@
...
+		return 0, 0, ErrExpectedNodeNotFound
+	}
+	start = int(stmt.If) + 2
+	end = latestEnd(start, stmt.Init, stmt.Cond)
+	return start, end, nil
+}
+
+type ForExtractor struct{}
+
+func latestEnd(start int, nodes ...ast.Node) (end int) {
```
joerdav, line 79: Nit: This function could be better placed above or below all the extractors.

parser/v2/goexpression/parse.go
```diff
@@ -0,0 +1,257 @@
...
+	// If the expression ends with `...` then it's a child spread expression.
+	if suffix := content[end:]; strings.HasPrefix(suffix, "...") {
+		end += len("...")
+	}
+	return start, end, nil
+}
+
+type IfExtractor struct{}
+
+var InGoIfExpression = false
```
joerdav, line 65: Is this used?

parser/v2/templelementparser.go
```diff
@@ -32,11 +13,10 @@ func (p templElementExpressionParser) Parse(pi *parse.Input) (n Node, ok bool, e
 		return
 	}
 
-	// Parse the identifier.
 	var r TemplElementExpression
-	if r.Expression, ok, err = templElementStartExpression.Parse(pi); err != nil || !ok {
-		err = parse.Error("templ element: found start '@' but expression was not closed", pi.Position())
-		return
+	// Parse the Go expresion.
+	if r.Expression, err = parseGo("if", pi, goexpression.Expression); err != nil {
```
joerdav, line 18:
```suggestion
	if r.Expression, err = parseGo("templ element", pi, goexpression.Expression); err != nil {
```

joerdav (commented):
This looks awesome, big fan of the go parse tests, should be really easy to add new ones without it becoming a huge unreadable file!

Just added a couple nits and some typos I spotted.

parser/v2/goexpression/parse.go
a-h: Good idea. Looking at it again, I've decided to remove the extractor types completely and inline the functions. It increases the locality of behaviour which I think makes it easier it reader.

PR #462, opened 2024-01-25
templ fmt: Add a way to emit to stdout even when reading from the filesystem

pedramos:
This seems to be the default behavior on other formatters (gofmt for example).

I'm interested in this change mostly because I use acme and acmego auto-formatter expects this behaviour. I expect the change to be useful in general though since the change reflects other formatters behaviour. 

I also added a warning to arg parsing, it seems only the first argument is used so now it issues a warning about the remaining arguments being discarded.

### DIFF
--- a/cmd/templ/fmtcmd/main.go
+++ b/cmd/templ/fmtcmd/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/a-h/templ/cmd/templ/processor"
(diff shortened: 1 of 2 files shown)


a-h:
It looks like this would break people's scripts if they're relying on the existing behaviour, which I don't want to do.

The existing behaviour is:

`templ fmt .` - formats all files in directory, in place
`templ fmt file.templ` - overwrites file.templ with formatted version
`templ fmt < file.templ` - writes formatted file.templ to stdout

I'm not sure why you can't use the `<` operator or similar in your environment to pipe the content, but if that's a hard limitation of your text editor, then I guess we could add a new boolean flag of `-stdout` which sets all output to be via stdout, with no files being written at all.

So, the existing behaviour would stay, and the new behaviour would be:

* `templ fmt -stdout file.templ` - writes the formatted result of file.templ to stdout
* `templ fmt -stdout .` - writes the formatted result of the director (and subdirs) to stdout (useful for checking that nothing needed formatting, by piping the input and the result to a hash function)
* `templ fmt -stdout < file.templ` - writes formatted file.templ to stdout (no change)

pedramos:
For clarification, I use acme as text editor and acmego for formatting, both live here https://github.com/9fans/go. 

Acmego reads from a file, but expects the result to be sent to stdout. This combination is not currently present in templ fmt. The -stdout is also ok to me, I can add that to the commit. 

I will do it when I have the time.

a-h:
Nice! Thanks for that. I'll hang on the flag.

pedramos:
hi @a-h ,
Changed the code so much I did a force push. It now has the -stdout flag and discussed. Please review and let me know if something is not ok.

a-h:
I made some updates to refactor it, and used the new logger. I had to make a new PR because I couldn't push to your branch, but you're listed as a contributor, since I branched from your PR.

Screenshot 2024-02-04 at 15 18 45

Thanks!

Issue #456, opened 2024-01-23
docs: Gin example cannot use context.


SebastianStehle:
If I understand it properly the gin example does not use the context, because there is no solution to pass the context from gin to the renderer.

I think there would be an alternative solution by using the `Render()` method of the gin context.

a-h:
Thanks, you're right!

I'd happily take a PR for an example update if you feel like it.

iamajoe:
https://github.com/a-h/templ/pull/466
You mean like this? Or are you thinking on something different?

a-h:
Closed in https://github.com/a-h/templ/commit/b6e7a779dacbddc4a185023be181fe7349e6b923

Thanks for your contributions!

PR #466, opened 2024-01-27
refactor(#456): gin example with context

iamajoe:
Augments gin example with a way to pass the gin request context to the render.


examples/integration-gin/main.go
```diff
@@ -14,5 +14,13 @@ func main() {
 		c.HTML(http.StatusOK, "", Home())
 	})
 
+	r.GET("/with-ctx", func(c *gin.Context) {
+		c.Render(http.StatusOK, &TemplRender{
```
SebastianStehle, line 18: I would create a function for that.

iamajoe: I would probably do it as well on my code but being an example, it seems that the abstraction could confuse people. Leaving it "open" and without that extra layer seems to me a better approach. It is easy enough to understand. That said, if the maintainers prefer that function in, I can set it up.

a-h: I'm not a Gin expert, I've never actually used it. I have no idea what the best approach is here.

I agree with the concept of keeping things simple in example code though.

@SebastianStehle - do you have an example of what you're talking about? I don't understand what you mean.

Issue #465, opened 2024-01-27
fmt: Ordered Html attributes


jonerer:
I really love that templ helps me structure and format the templ files, like how putting a newline between some Html attributes turns the whole div into a multiline div and lays out the attributes on one line each. 

I wish that templ would take this one step further and automatically sort the attributes. Probably by alphabetical order, but maybe there's some other common sorting algorithm. It doesn't matter that much, as long as it's consistent. 

The point being that having attributes in a consistent order would really simplify the readability/skimmability of templ files, and that doing this manually is too much work. For instance if you know that an <input /> will always be in the same ordering like "class", "name", "id" "type", "value", it would make it much easier to skim the code. For example to find where things might be wrong, what's missing and where to add things. Many times I've accidentally added two different "class" attributes, for instance. 

Is there consensus that this would be an improvement? Of course it will have an effect on everybody's code style, which some people might not like. But in generanl I think people are typically in favor of having consistency and automation over having their own specific style. 

I think making the change could probably be done around here: 
https://github.com/a-h/templ/blob/c683be18add8e2f6465c9ab6044f7360cf50aed3/parser/v2/types.go#L663

EDIT:
Desired behaviour:
```
# cat thing.templ
...
<input type="text" class="mycls" name="weo" value="myvalue"/>
<input name="wuw" value="thing" class="input" type="text"/>
<input name="weow" type="text" class="input input-bordered"/>
...
# templ fmt
# cat thing.templ
...
<input class="mycls" name="weo" type="text"  value="myvalue "/>
<input class="input" name="wuw" type="text" value="thing" />
<input class="input input-bordered" name="weow" type="text" />
....
# Aaah, so neat and tidy
```

jonerer:
I made an initial PR. I could update it if you prefer another code style. There are two questions:

- How do you want the "Name" attribute to be found? Either by casting to known Attribute structs (like ConstantAttribute), by reflection (as done) or by adding a "Name()" method to the Attribute struct
- Do you have an opinion on where conditional attributes get put? As of the first implementation they're just put last

joerdav:
Thanks for the suggestion. This could definitely be useful, and you've identified where this change should live.

One thing first though is whether this is going to be a benefit or an annoyance for most people. I think I'm on your side of the fence, but I think it could be frustrating if you want your attrs in a specific order.

Could be worth taking notes from other HTML formatters on this. Are there any that do this as a default behavior, is it a big point of contention for them, or does everyone like it?

[...]

PR #469, opened 2024-01-28
feat: Sort html attributes

jonerer:
For discussion see #465


joerdav:
I forgot to close this when the discussion came to a close! There is some work to open up to other tooling like different HTML lsps, so maybe this will be an option in the future!

Issue #439, opened 2024-01-19
cmd: clean up generate stdout/stderr


egtann:
First off, thank you for this amazing tool. I am loving the experience with templ so far! I do have some small ideas which would improve quality of life when using `templ generate`.

There's two small improvements to the UX of the CLI tool that we can make:

### 1, stdout
Each time `templ generate` runs, it sends a very large amount of text to stdout on success. This is at odds with how most Unix and Go tools work, since those generate no output on success. Ideally `templ generate` would behave like other tools and not output hundreds of lines on every build in a large project, as this effectively wipes my terminal on each run given how many lines it outputs.

If we want to output debug information like how long every file took to generate, that should be under a verbose flag.

To work around this, I've started using `templ generate 1>/dev/null` in a Makefile, but this leads me to my next issue.

### 2. stderr
Failures/errors are outputted to stdout, so when I get an error, if I'm discarding the noise from stdout, I get no output or an unhelpful error:

```
$ templ generate 1>/dev/null
# <there's no output if it errors>

$ make templ
make: *** [Makefile:2: templ] Error 1
```

a-h:
I've introduced a new structured logger, with a custom colourised handler in https://github.com/a-h/templ/pull/470 while I was rewriting everything.

I thought that a fancy TUI was overkill once the output was tidied up. By default, it logs at `Info` level, so the default output is much simpler, just providing a summary.

```
$ templ generate
(✓) Complete [ updates=54 duration=87.295806ms ]
```

It's now under a verbose flag. You can set `-v` for debug mode, or set the level to "warn", "error" etc.. with the `-log-level` flag if you want no output at all unless there's errors.

However, I've kept the logs on stdout, since go's slog handler only has the choice of one and the logs *are* the program output.

a-h:
Update on this one. Decided to log to `stderr`, since it affects Bazel and other users to log to `stdout`.

https://github.com/a-h/templ/issues/753

Issue #395, opened 2024-01-04
templ generate hangs


nexovec:
Templ generate generates code but doesn't exit... this happens:

```bash
templ generate
Processing path: /home/nexovec/src/app
(✓) Generated code for "/home/nexovec/src/app/internal/templates/auxiliary.templ" in 8.737188ms
(✓) Generated code for "/home/nexovec/src/app/internal/templates/articles.templ" in 9.527104ms
(✓) Generated code for "/home/nexovec/src/app/internal/templates/common.templ" in 12.512291ms
(✓) Generated code for "/home/nexovec/src/app/internal/templates/app.templ" in 14.644502ms
(✓) Generated code for 4 templates with 0 errors in 15.006609ms
```
and here it freezes up.

This started happening yesterday. templ version: `v0.2.513`, `go version go1.21.5 linux/amd64`

NOTE: I also had trouble migrating to the newest version, was getting errors about a bunch of symbols of templ not being found, but I regenerated everything and it works now. I think I just messed it up somehow myself, just saying in case this reoccurs somewhere else.

dougbarrett:
This is also happening to me as well - I'm running templ v0.2.513 and go version go1.21.5 darwin/arm64.

I've stripped it down to a basic template file and am consistently replicating the issue:

```
package components

templ layout() {
    <div>
        hi
    </div>
}
```

a-h:
Interesting.... Looks like we've got one example of Linux and another on Mac. Thanks for the info.

I wasn't able to replicate the issue on my ARM64 Mac:

```
1 0 /Users/adrian/github.com/a-h/templ/generator/test-395 % which templ
templ not found
1 1 /Users/adrian/github.com/a-h/templ/generator/test-395 % go install github.com/a-h/templ/cmd/templ@latest
1 0 /Users/adrian/github.com/a-h/templ/generator/test-395 % templ generate
Processing path: /Users/adrian/github.com/a-h/templ/generator/test-395
(✓) Generated code for "/Users/adrian/github.com/a-h/templ/generator/test-395/template.templ" in 1.682083ms
(✓) Generated code for 1 templates with 0 errors in 1.878583ms
1 0 /Users/adrian/github.com/a-h/templ/generator/test-395 %
```

<img width="1512" alt="Screenshot 2024-01-05 at 09 24 05" src="https://github.com/a-h/templ/assets/1029947/5aa2cc57-4423-4714-a22a-aa1d2144bcc3">

[...]

PR #470, opened 2024-01-28
refactor: complete rewrite of the generation command, adds fsnotify hot reload

a-h:
This PR updates the generation command.

This could solve various issues, including:

* github.com/a-h/templ/issues/464 - since the command execution code has been rewritten
* https://github.com/a-h/templ/issues/439 - I've added slog, with a custom colourised handler
* https://github.com/a-h/templ/issues/395 - since the code has been completely rewritten
* https://github.com/a-h/templ/issues/146 - since the hot reload code is completely new and now uses fsnotify

This builds on top of the excellent PR from @stephenafamo (https://github.com/a-h/templ/pull/366) which added the dev mode facility where Go recompilation isn't required if only the text changes.

Here's a video showing the updated experience:

https://github.com/a-h/templ/assets/1029947/70fbfe60-2197-4095-ad1c-fd59971a5386


joerdav:
Still in the process of grokking the bulk of this PR! An initial comment would be on the new `level` flag, I wonder if a more standard flag approach such as `-v` and `-vv` would be more intuitive.

FACorreiaa:
Im guessing this will also work if Im running the app from a Docker container for testing ? Assuming it can be used with air.

joerdav:
@FACorreiaa I don't think this would interop with air too well, it's almost an alternative. Air does the watching for you, and will rebuild your code anyway if I understand it correctly. 

I think we would have to have a mode where templ only regenerates templ code, and then you could leave the recompiling of Go when required to Air.

a-h:
If you run `templ generate --watch` it generates dev mode code which - creates/updates `_templ.txt` files with the strings that make up the HTML, and Go code that reads from the file instead of having the strings embedded in the output code.

If you get `air` to run `templ generate --watch` continuously, that would be ideal (if you want to use air to do other things). The key thing is that if you have `templ generate --watch` mode running, `air` shouldn't restart your program if `*.templ` or `_templ.txt` files are updated, but should if any `*.go` files are updated.

If you run `templ generate --watch --cmd=""`, then templ is smart enough to only stop and start the cmd if that's the case.

If you then add `templ generate --watch --cmd="go run ." --proxy="http://localhost:8080"` templ starts a proxy and opens the browser to that page. The proxy adds some JS to the page that initiates a background SSE (server-sent events) session. If the text on a page is updated, or the Go code changes and the cmd has to reload, the proxy mode then sends an SSE to the browser triggering a page reload (i.e. you don't need to reload the web browser page manually to see the changes).

cmd/templ/sloghandler/handler.go
```diff
@@ -0,0 +1,105 @@
...
+	"log/slog"
+	"strings"
+	"sync"
+
+	"github.com/fatih/color"
+)
+
+var _ slog.Handler = &Handler{}
+
+type Options struct {
```
joerdav, line 15: It doesn't look like this wound up being used.

joerdav:
Looks good to me!

stefanohrmann:
https://github.com/a-h/templ/blob/1a1ba9731926fe6382bb45d4d9c773ad75632a3c/cmd/templ/generatecmd/cmd.go#L252-L255

Where is this relevant? I didn't see any code that pushes `ctx.Err()` to the `errs` channel.

a-h:
> https://github.com/a-h/templ/blob/1a1ba9731926fe6382bb45d4d9c773ad75632a3c/cmd/templ/generatecmd/cmd.go#L252-L255
> 
> Where is this relevant? I didn't see any code that pushes `ctx.Err()` to the `errs` channel.

I don't think there is any code that does that. I think I can remove it.

Issue #387, opened 2024-01-02
lsp: Support goToDefinition from a Go file to a templ file


guettli:
If I ctrl-click on `page()` then ...

![image](https://github.com/a-h/templ/assets/414336/37f0e991-0125-4609-8ced-2f9061561442)

then I get to the autogenerated Go code:

![image](https://github.com/a-h/templ/assets/414336/7f927b64-7fbe-4eae-8977-7f0fcbcfdbbc)

It would be great if I could get to the templ-file instead:

![image](https://github.com/a-h/templ/assets/414336/d418e9f7-e5c9-4f00-b36a-9359e9cf649f)

a-h:
Yes, that would be much more useful!

If you're in a `*.templ` file, then the templ LSP is operating and will carry out a remapping operation from templ position -> go file position -> gopls -> templ position, as per here: https://github.com/a-h/templ/blob/c674f647606fe180063d074ff12720ec1dd92350/cmd/templ/lspcmd/proxy/server.go#L507-L512

But currently, if you start from a `*.go` file, then `gopls` is operating and there's no templ code running at all. To rewrite the responses from `gopls`, templ would have to proxy all Go LSP operations (i.e. for all `*.go` files) and work out that if a `_templ.go` file was being referenced, then, it should look up the various source code mapping positions and switch it to `*.templ` instead. I'm not sure if replacing (as a proxy) gopls for all projects is ideal (higher risk?), but it would work.

It may also possible for the templ LSP to capture `*_templ.go` file usage too, and be able to ctrl+click on the function name to get to the `*.templ` file. (Worse UX, but maybe a good first step)

joerdav:
I know you can have multiple LSPs attached to a file, but I wonder what would happen if you had 2 trying to respond to a jump to definition request.

guettli:
I asked the gopls maintainers. Maybe they know a hook which templ could implement: https://github.com/golang/go/issues/65001

joerdav:
We have the following issue to implement the line directive https://github.com/a-h/templ/issues/415, and then I think it's a case of closing this ticket and helping the gopls team implement https://github.com/golang/go/issues/65001

guettli:
@joerdav thank you for your work! This is great!

joerdav:
Thank @af-md !

meistertigran:
Here's a quick and dirty fix for anyone using neovim. It doesn't go directly to the location of the definition in the file, but at least it opens the correct file. 
```lua
[...]

PR #476, opened 2024-01-29
feat: Implement the go line compiler directive

af-md:


joerdav:
I think we didn't understand the other impacts that this change would have fully, I think we should close this for now and maybe revisit some other options.

embiem:
@joerdav could you please share what these impacts are? I'm looking into the lsp goToDefinition [issue](https://github.com/a-h/templ/issues/387) again and currently it seems like this is unresolved and not actively worked on. I'm willing to dig into this.

The discussion in the [related go issue](https://github.com/golang/go/issues/65001) is also open with last status being that it's waiting on this line directive addition, which is now closed.

joerdav:
Absolutely. The main issue is that the go line directive affects the debug information at runtime for code, my concern was that I actually don't know what else this would affect.

I know that for any panics originating from a templ components the stack traces would have line numbers that are pointing to completely the wrong place, and be missleading.

If we were happy with this and knew what other effects there would be, then I think I'd be comfortable merging, because this issue though small has quite a high demand.

PR #477, opened 2024-01-30
chore: watch tests - give 5 attempts for the server to start

joerdav:


joerdav:
This test consistently failed on my slow intel mac.

PR #484, opened 2024-02-01
Css args

### LINKED ISSUE #88: css: Support arguments in CSS components
I am attempting build dynamic pie charts using CSS where the percentage would be controlled by a CSS variable calculated on the server. I'm not sure if this can currently be achieved in templ because style attributes do not allow for templ expressions.

joerdav:
Fixes #88


parser/v2/templateparser.go
```diff
@@ -28,7 +28,7 @@ var templateExpressionParser = parse.Func(func(pi *parse.Input) (r templateExpre
 	// templ (x []string) Test() {
 	// becomes:
 	// func (x []string) Test() templ.Component {
-	if r.Expression, err = parseGoFuncDecl(pi); err != nil {
+	if _, r.Expression, err = parseTemplFuncDecl(pi); err != nil {
```
a-h, line 31: I wonder if there's some debug value in the name of the templ, it could be passed down the context.

For example, the context could identify slow rendering components, or could print out a series of names: `Page > Body > content > heading`.

Not important, or that you need to do anything with this PR. Its just that your change sparked ideas. 😀

a-h (approved):
Brilliant. Nice to see this long requested feature implemented.

I'm happy to see the new Go parser paying off too. 😀

jonerer:
Looks very nice! The example in the documentation showcases an integer parameter, does it also handle strings?

Like for example background url('{ myVar }'). If so, what kind of escaping is used?

koddr:
It looks exactly the way I need it to look! 👍 

@a-h when are you planning to release this?

a-h:
Plan to release tomorrow.

PR #496, opened 2024-02-06
fix: stop losing diagnostics from nested elements

### LINKED ISSUE #478: parser: legacy syntax does not produce a warning from within if/for/switch
The parser has been expanded to warn about the use of the legacy template call syntax `{! component() }`, the warning happens both in the lsp, and the generate command. And the format command auto fixes these issues.

However, the diagnostic does not surface if the syntax is in an if statement. Same goes for for/switch statements.

```
package main

templ index() {
  if true {
    {! templ.Raw("") }
  }
}
```

Create the above in a templ file, run `templ generate`.

Expected: a warning message is displayed.

Actual: no warning message is shown.

joerdav:
Fixes #478

### DIFF
--- a/parser/v2/templateparser.go
+++ b/parser/v2/templateparser.go
@@ -108,6 +108,9 @@ func (p templateNodeParser[T]) Parse(pi *parse.Input) (op Nodes, ok bool, err er
 					Range:   n.Expression.Range,
 				})
 			}
+			if n, ok := node.(DiagnosticNode); ok {
+				op.Diagnostics = append(op.Diagnostics, n.GetDiagnostics()...)
+			}
 			if matched {
 				op.Nodes = append(op.Nodes, node)
 				break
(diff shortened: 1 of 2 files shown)


joerdav:
@a-h am I okay to merge this? Just to make the warning more consistent, and provide a better interface for surfacing diagnostics.

a-h:
I was going to ask about the bubbling up of diagnostics, and whether that was intentional, or if the diagnostics should be brought out by "walking" the template file.

It looked like you could end up with duplicate diagnostics...

joerdav:
I'm not sure you could end up with duplicates, you could end up with the same diagnostic for different lines of code, but I think that's useful, as the log message prints the line/col numbers.

joerdav:
I'll try a refactor, I had thought maybe a `type Diagnoser func(Node) Diagnostic` could be nice. I wanted to avoid walking the whole tree but for simplicity it may be best, without having to change anything in `types.go`.

Issue #454, opened 2024-01-23
parser: Allow exception of unsafe style attribute expressions.


chrisbward:
Unsure how to proceed here - as there's no mention in the docs regarding the style attribute, but an old ticket mentioning templ.RawCSS, and I wasn't able to get this to work - thanks!

iamajoe:
You can go around with templ.Attributes.

https://github.com/a-h/templ/blob/main/parser/v2/types.go#L483
This is done for security purposes from my understanding, right templ team?

I've also got this issue and I am wondering though, how much of an issue it is to actually supoprt it?! What could be done? I don't mind working on this but I don't want to go against the direction of the templ team.

joerdav:
Correct, this is a security control. But it does bring into question whether it should be escaped, similar to how we escape URL sanitisation.

iamajoe:
In my opinion, I think there should be an way to escape. I can give an example where using style makes sense.

Lets say you have a loading bar of some sort. For onboarding for example. In general, the best way would be to setup a `width: 74%` but the "only" way for you to do that would be through `style`. You can't use the `css` as well because this value needs to stay dynamic.

This situation has happened to me several times while using templ. My solution was to use templ.Attributes but that was just a way for me to circumvent the safeties. 

Style security becomes a little bit shaded when you think that you have full control over your code and what goes into style. To "ban" the usage on simple cases because of a very specific case seems a bad move?! That said, I like the idea of escape with sanitisation like it is done with URL.

joerdav:
Would the proposal on the following help you, it's a slightly different solution: https://github.com/a-h/templ/issues/88

iamajoe:
Ah yes, where you pass the argument to css? Definitely. I actually thought was possible until I figured that couldn't be done with current templ.

I still think that denying style is a bit of an opinionated option that removes flexibility out of the html usage. Even though I like your proposal, I still think that it is a bandaid (a cool and good one) around the denial of an attribute of the html standard.

I think your proposal and the the sanitize of style like it is done with URL would be invaluable. Just my 2 cents.

chrisbward:
+1 for supporting inline safely

chrisbward:
> In my opinion, I think there should be an way to escape. I can give an example where using style makes sense.
> 
> Lets say you have a loading bar of some sort. For onboarding for example. In general, the best way would be to setup a `width: 74%` but the "only" way for you to do that would be through `style`. You can't use the `css` as well because this value needs to stay dynamic.
> 
[...]

PR #534, opened 2024-02-18
docs: Add JSON attribute example

MarhyCZ:
Hi, based on our discussion in #454, I added an example into docs on how to pass JSON object as HTML attribute value. I see that meanwhile same question already cropped up in #519. I stick to the more generic usecase so that it nudges the reader in the right way.

### DIFF
--- a/docs/docs/03-syntax-and-usage/03-attributes.md
+++ b/docs/docs/03-syntax-and-usage/03-attributes.md
@@ -160,3 +160,20 @@ templ Button(text string) {
 ## CSS attributes
 
 CSS handling is discussed in detail in [CSS style management](css-style-management).
+
+## JSON attributes
+
+If you want to pass any JSON object as an attribute value, you need to serialize it into a string.
+
+```go
+func countriesJSON() string {
+	countries := []string{"Czech Republic", "Slovakia", "United Kingdom", "Germany", "Austria", "Slovenia"}
+	bytes, _ := json.Marshal(countries)
+	return string(bytes)
+}
+```
+```templ
+templ SearchBox() {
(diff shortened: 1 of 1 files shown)


docs/docs/03-syntax-and-usage/03-attributes.md
```diff
@@ -160,3 +160,20 @@ templ Button(text string) {
...
+
+If you want to pass any JSON object as an attribute value, you need to serialize it into a string.
+
+```go
+func countriesJSON() string {
+	countries := []string{"Czech Republic", "Slovakia", "United Kingdom", "Germany", "Austria", "Slovenia"}
+	bytes, _ := json.Marshal(countries)
+	return string(bytes)
+}
+```
```
a-h, line 174:
```suggestion
```

```

File: docs/docs/03-syntax-and-usage/03-attributes.md [hunk 2]
```diff
@@ -160,3 +160,20 @@ templ Button(text string) {
 ## CSS attributes
 
 CSS handling is discussed in detail in [CSS style management](css-style-management).
+
+## JSON attributes
+
+If you want to pass any JSON object as an attribute value, you need to serialize it into a string.
```
a-h, line 166:
```suggestion
To set an attribute's value to a JSON string (e.g. for HTMX's [hx-vals](https://htmx.org/attributes/hx-vals) or Alpine's [x-data](https://alpinejs.dev/directives/data)), serialize the value to a string using a function.
```

a-h (approved):

a-h:
Thanks for your contribution here. Improving the docs is always a good thing! 😁

PR #546, opened 2024-02-22
feat: add fuzzing suite to expression parser

a-h:


joerdav:
Cool to see the things it's already found and the fixes you've implemented already!

Not really a note on your PR particularly, but just food for thought, I wonder how we can implement this into CI? We don't want it running on every commit...

Maybe could somehow fit into the process of releasing a new version? The CI runs the fuzz for a while before actually tagging the release?

a-h:
Go fuzzing is nicely thought out, it re-runs discovered crashes when you run go test, but apart from that, you've got to explicitly run the fuzzing.

I'm not sure of the value of running it a lot, since we only need to run it if the goexpression package changes.

joerdav:
Fair, I suppose you could have it run when a new tag is pushed, only if that package is updated. Agree it should not be ran frequently.

Issue #498, opened 2024-02-07
lsp: Add HTML Intellisense Support to Templ LSP


salihdhaifullah:
Currently, the Templ LSP lacks native support for HTML language features. This feature request is to propose the addition of HTML LSP support to the Templ LSP.

joerdav:
Hi @salihdhaifullah thanks for the proposal.

I just wanted to check if you've tried out the HTML options that are outlined in the docs here: https://templ.guide/commands-and-tools/ide-support

I'm using Neovim and have enabled the HTML LSP on templ files and it works perfectly.

If you're using vscode then the option is to enable emmet completion.

If these are not sufficient then can you please reopen this ticket with details of what you found was lacking?

yardenshoham:
I think they mean something like 
![image](https://github.com/a-h/templ/assets/20454870/480697ec-8f37-4731-9576-1c6f41d7be50)

It doesn't work in templ
![image](https://github.com/a-h/templ/assets/20454870/af826605-6e86-479c-a015-b725617338b7)

joerdav:
I've marked the issue as needing some investigation, so we can explore options before taking this as feature for the templ LSP.

joerdav:
Unfortunately, I don't think there is a solution for this. Go templates has the exact same problem in vscode, either you edit the file as a gotmpl or as a html file:

https://github.com/golang/vscode-go/wiki/features#go-template-syntax-highlighting

yardenshoham:
Can the templ LSP proxy the HTML LSP?

joerdav:
I think @a-h had some thoughts on this. But I can see proxying another LSP as a pain, it's really frustrating that vscode currently doesn't support multiple file associations. 

Often the templ LSP has trouble finding gopls to run and proxy, I think having more executables required for templ LSP would damage the ease of install and config. I guess an option would be for proxying the HTML LSP to be an optional feature if the LSP is found, but again balancing multiple LSPs should be the role of the client as Neovim does.

yardenshoham:
Gotcha. I chose templ because I thought it supports this and that (for me) was the "wow" effect. From https://templ.guide/:
> Great developer experience: Ships with IDE autocompletion.

I assumed that meant HTML as templ is for HTML.

joerdav:
Ah, no that's mostly for the Go side of it, type safety and autocomplete of Go expressions.

a-h:
[...]

PR #548, opened 2024-02-23
feat: parse HTML element and attribute positions

alehechka:
# Overview
This PR adds positions to the templ parser results for HTML elements and attributes relating to conversation in  https://github.com/a-h/templ/issues/498, specifically the recommendation made in https://github.com/a-h/templ/issues/498#issuecomment-1952638089.

This handles only calculating expression positions. I plan to open another PR when ready to use the calculated positions to enable LSP intellisense with a few hard-coded results. After that, I can work towards the solution I recommended in https://github.com/a-h/templ/issues/498#issuecomment-1952566968, to import an HTML LSP's data into templ's LSP.


parser/v2/types.go
```diff
@@ -405,6 +405,17 @@ type WhitespaceTrailer interface {
...
+type HTMLExpression interface {
+	HTMLExpression() Expression
+}
+
+var (
+	_ HTMLExpression = Element{}
+	_ HTMLExpression = BoolConstantAttribute{}
+	_ HTMLExpression = BoolExpressionAttribute{}
+	_ HTMLExpression = ConstantAttribute{}
+)
```
alehechka, line 417: I added this with plans to eventually use the interface to determine when the HTML LSP can be enabled for intellisense.

alehechka:
@a-h, I finished getting all the unit tests updated to reflect the position calculations so this should be ready for code review now.

joerdav:
My immediate thoughts here are that maybe `Expression` isn't the right type, so far I think we only store Go in here.

I could be being nit-picky but I think it makes more sense for there to be an HTML specific expression type.

alehechka:
@joerdav, that sounds valid. I mainly reused it since it was already a pattern for storing positions. Should be simple enough to do a find and replace with a new name. Do you have anything in mind as an alternative to `Expression`?

alehechka:
Considering that `Expression.Value` ends up with the same value as `Name` from the parent struct in all of these cases, we could probably just use `Range` directly to store the positions of elements/attributes.

alehechka:
@joerdav, I went ahead and replaced `Expression` with `Range` in all cases. Let me know if this approach works better with you.

joerdav:
@alehechka the type range makes sense, happy with that. I wonder if a better name for the property could be in order? The name `Range Range` to me suggests it's the range of the whole open tag, maybe `NameRange Range` could be more specific? After that I'm happy with this PR if @a-h is.

a-h (approved):

PR #550, opened 2024-02-23
docs: fix gin html status code overwrite and provide fallback renderer

Foggyman:
Refactoring to better integrate with gin.HTML function (allow different http status code) and optionally add fallback html renderer to gin HTMLRender if no temp component is provided

### DIFF
--- a/examples/integration-gin/main.go
+++ b/examples/integration-gin/main.go
@@ -9,7 +9,12 @@ import (
 
 func main() {
 	engine := gin.Default()
-	engine.HTMLRender = gintemplrenderer.Default
+	engine.LoadHTMLFiles("./home.html")
+
+	//engine.HTMLRender = gintemplrenderer.Default
+
+	ginHtmlRenderer := engine.HTMLRender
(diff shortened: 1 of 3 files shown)


examples/integration-gin/main.go
```diff
@@ -9,7 +9,12 @@ import (
 
 func main() {
 	engine := gin.Default()
-	engine.HTMLRender = gintemplrenderer.Default
+	engine.LoadHTMLFiles("./home.html")
+
+	//engine.HTMLRender = gintemplrenderer.Default
```
a-h, line 14: This seems like it's been left behind.

examples/integration-gin/main.go
```diff
@@ -9,7 +9,12 @@ import (
...
+
+	ginHtmlRenderer := engine.HTMLRender
+	engine.HTMLRender = &gintemplrenderer.HTMLTemplRenderer{FallbackHtmlRenderer: ginHtmlRenderer}
```
a-h, line 17: This doesn't feel right. I don't see why the variable exists.

Foggyman: the variable can be removed indeed, it was to make the code a little more readable so that user can understand that here we are setting a fallback to gin default html renderer in case we do not pass a templ

PR #552, opened 2024-02-23
chore: add gzip test for proxy feature

JamesTiberiusKirk:
NGL, was a bit confused reading how the tests worked, so feel free to suggest a differnt way.

Basically I just duplicated the `TestFileModificationsResultInSSE` test and modified the `Setup()` func to configure the test server with gzip flag.


a-h:
For some reason, couldn't push the rebased version to your branch.

Thanks for the test, great stuff! I also added a unit test alongside.

Issue #525, opened 2024-02-14
cmd: templ generate with watch can generate corrupted Go files, because of race condition with fsnotify WRITE events


stefanohrmann:
I used the following command:

```sh
templ generate -watch -v
```

to watch my template directory and to generate the Go files on changes. In the first place I used the command with the `proxy` and the `cmd` option to automatically update the displayed website, but the issue also occurs if I only use the the `watch` option. I got the following trace:

```sh
stefan@home:~/Projekte/go/watchlist$ templ generate -watch -v
(!) templ version check: generator v0.2.570 is newer than templ version v0.2.543 found in go.mod file, consider running `go get -u github.com/a-h/templ` to upgrade
(✓) Starting post-generation handler
(✓) Walking directory [ path=/home/stefan/Projekte/go/watchlist devMode=true ]
(✓) Starting event handler
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/index_templ.go op=CREATE fileSize=4238 ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/index.templ op=CREATE fileSize=1037 ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/index_templ.txt op=CREATE fileSize=675 ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout.templ op=CREATE fileSize=536 ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout_templ.txt op=CREATE fileSize=444 ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout_templ.go op=CREATE fileSize=1386 ]
(✓) Watching files
(✓) Waiting for context to be cancelled to stop watching files
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout_templ.go op=WRITE fileSize=1386 ]
(✓) Generated code [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout.templ in=1.693935ms ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout_templ.go op=WRITE fileSize=1386 ]
(✓) Processing file [ file=/home/stefan/Projekte/go/watchlist/ui/web/templates/layout_templ.txt op=WRITE fileSize=444 ]
[...]

PR #555, opened 2024-02-24
feat: debounce generate watch events

joerdav:
Fixes: #525

### DIFF
--- a/cmd/templ/generatecmd/watcher/watch.go
+++ b/cmd/templ/generatecmd/watcher/watch.go
@@ -104,9 +104,6 @@ func (w *RecursiveWatcher) loop() {
 			if !ok {
 
 				t = time.AfterFunc(100*time.Millisecond, func() {
-					w.timerMu.Lock()
-					delete(w.timers, event.Name)
-					w.timerMu.Unlock()
 					w.Events <- event
 				})
 				w.timerMu.Lock()
--- a/cmd/templ/generatecmd/watcher/watch.go
+++ b/cmd/templ/generatecmd/watcher/watch.go
@@ -72,8 +72,8 @@ type RecursiveWatcher struct {
 	w       *fsnotify.Watcher
 	Events  chan fsnotify.Event
 	Errors  chan error
-	timers  map[string]*time.Timer
 	timerMu sync.Mutex
+	timers  map[string]*time.Timer
 }
 
 func (w *RecursiveWatcher) Close() error {
--- a/cmd/templ/generatecmd/watcher/watch.go
+++ b/cmd/templ/generatecmd/watcher/watch.go
@@ -6,11 +6,18 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/fsnotify/fsnotify"
 )
 
-func Recursive(ctx context.Context, path string, out chan fsnotify.Event, errors chan error) (w *RecursiveWatcher, err error) {
+func Recursive(
+	ctx context.Context,
+	path string,
+	out chan fsnotify.Event,
+	errors chan error,
+) (w *RecursiveWatcher, err error) {
 	fsnw, err := fsnotify.NewWatcher()
 	if err != nil {
 		return nil, err
@@ -20,6 +27,7 @@ func Recursive(ctx context.Context, path string, out chan fsnotify.Event, errors
 		w:      fsnw,
 		Events: out,
 		Errors: errors,
+		timers: make(map[string]*time.Timer),
 	}
 	go w.loop()
 	return w, w.Add(path)
@@ -60,10 +68,12 @@ func shouldIncludeFile(name string) bool {
 }
 
 type RecursiveWatcher struct {
-	ctx    context.Context
-	w      *fsnotify.Watcher
-	Events chan fsnotify.Event
-	Errors chan error
+	ctx     context.Context
+	w       *fsnotify.Watcher
+	Events  chan fsnotify.Event
+	Errors  chan error
+	timers  map[string]*time.Timer
+	timerMu sync.Mutex
 }
 
 func (w *RecursiveWatcher) Close() error {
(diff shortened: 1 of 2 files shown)


cmd/templ/generatecmd/watcher/watch_test.go
```diff
@@ -0,0 +1,106 @@
...
+	rw.loop()
+	count := 0
+	exp := time.After(300 * time.Millisecond)
+	for {
+		select {
+		case <-rw.Events:
+			count++
+		case <-exp:
+			if count != 2 {
+				t.Errorf("expected 1 event, got %d", count)
```
zolrath, line 68: expected 2 events

cmd/templ/generatecmd/watcher/watch_test.go
```diff
@@ -0,0 +1,106 @@
...
+	rw.loop()
+	count := 0
+	exp := time.After(500 * time.Millisecond)
+	for {
+		select {
+		case <-rw.Events:
+			count++
+		case <-exp:
+			if count != 2 {
+				t.Errorf("expected 1 event, got %d", count)
```
zolrath, line 101: expected 2 events

cmd/templ/generatecmd/watcher/watch_test.go
```diff
@@ -0,0 +1,106 @@
...
+		w: &fsnotify.Watcher{
+			Events: make(chan fsnotify.Event),
+		},
+		Events: make(chan fsnotify.Event, 2),
+		timers: make(map[string]*time.Timer),
+	}
+	go func() {
+		rw.w.Events <- fsnotify.Event{Name: "test.templ"}
+		<-time.After(200 * time.Millisecond)
+		rw.w.Events <- fsnotify.Event{Name: "test2.templ"}
```
zolrath, line 88: This should be `test.templ` if we're testing that the same file with a sufficient gap between events registers both events.

cmd/templ/generatecmd/watcher/watch.go
```diff
@@ -60,10 +68,12 @@ func shouldIncludeFile(name string) bool {
...
-	ctx    context.Context
-	w      *fsnotify.Watcher
-	Events chan fsnotify.Event
-	Errors chan error
+	ctx     context.Context
+	w       *fsnotify.Watcher
+	Events  chan fsnotify.Event
+	Errors  chan error
+	timers  map[string]*time.Timer
+	timerMu sync.Mutex
```
zolrath, line 76: I know the official go codebase places mutexes directly above the struct item they're protecting, so in this case you'd place the timerMu directly before timers. Not sure if that style is replicated here, but thought I'd mention it.

zolrath (commented):
A couple concrete things in the test suite along with a nit!

stefanohrmann:
I looked through your changes and I think you are debouncing every `fsnotify` event regardless of the `Op` type, that means a `REMOVE` with a `CREATE` shortly after will emit as a `REMOVE` event. The example for debouncing/deduping of `fsnotify` has a guard for only applying the debouncing/deduping to `CREATE/WRITE` events:

https://github.com/fsnotify/fsnotify/blob/c94b93b0602779989a9af8c023505e99055c8fe5/cmd/fsnotify/dedup.go#L80-L82

joerdav:
@stefanohrmann Interesting, seems the example you provided `REMOVE` would be skipped entirely. I wonder if we dedupe by only events with the same event?

cmd/templ/generatecmd/watcher/watch.go
joerdav: Didn't know of this, but will take your word on it :)

stefanohrmann:
> @stefanohrmann Interesting, seems the example you provided `REMOVE` would be skipped entirely. I wonder if we dedupe by only events with the same event?

Yes the example is really simple and I think a `REMOVE` event should clear the timer and prevent the emitting of any previous `CREATE/WRITE` event(s). Your current implementation does only check for a certain filename/op combination and not that another event for the same filename was already scheduled.

cmd/templ/generatecmd/watcher/watch.go
```diff
@@ -73,7 +73,19 @@ type RecursiveWatcher struct {
 	Events  chan fsnotify.Event
 	Errors  chan error
 	timerMu sync.Mutex
-	timers  map[string]*time.Timer
+	timers  map[timerKey]*time.Timer
+}
+
+type timerKey struct {
```
stefanohrmann, line 79: Do you plan to extent this structure? It has the same fields as `fsnotify.Event` and you could take the event itself as key.

joerdav: My only thinking was that this protects from a future where the fsnotify event gets a new field that we don't want to debounce on. What do you think?

cmd/templ/generatecmd/watcher/watch.go
```diff
@@ -88,7 +110,21 @@ func (w *RecursiveWatcher) loop() {
 			if !shouldIncludeFile(event.Name) {
 				continue
 			}
-			w.Events <- event
+			tk := timerKeyFromEvent(event)
+			w.timerMu.Lock()
+			t, ok := w.timers[tk]
+			w.timerMu.Unlock()
+			if !ok {
+
```
a-h, line 118:
```suggestion
```

cmd/templ/generatecmd/watcher/watch.go
a-h: I think your logic is sound there @joerdav. It reads well to me.

a-h (approved):
LGTM!

cmd/templ/generatecmd/watcher/watch.go
stefanohrmann: @joerdav Sounds good. I was just puzzled, because other parts of the watcher uses `fsnotify.Event` for notification.

Issue #561, opened 2024-02-26
docs: Inline method components


Ra0k:
The language supports method components like this:

```go
package main

type Data struct {
	message string
}

templ (d Data) Method() {
	<div>{ d.message }</div>
}

func main() {
	d := Data{
		message: "You can implement methods on a type.",
	}
	d.Method().Render(context.Background(), os.Stdout)
}
```

Would it be possible to add support for this kind of syntax?
Sorry, if it is already supported, I did not find anything about the specific syntax in the documentation nor a working example. 

```go
templ Hello() {
	@Data{
		message: "You can implement methods on a type.",
	}.Method()
}
```

This syntax could be a cleaner way of configuring renderable components than passing config arguments to functions.

Ra0k:
@joerdav this syntax does not work for me as it is. 

Extending the [demo code](https://templ.guide/core-concepts/components/#method-components) with the following:
```go
templ Hello() {
	@Data{
		message: "You can implement methods on a type.",
	}.Method()
}
```

will result in this error message:
`Data.Render undefined (type Data has no field or method Render`

Ra0k:
Thank you, @joerdav! I haven't realized that it's now supported in the new version! I can confirm that it indeed works.

PR #565, opened 2024-02-27
chore(docs): add inline component example

Ra0k:
Based on #561 

This PR adds missing "os" import to the existing method component example and add a new example for inline struct initialization and component method calling.

### DIFF
--- a/docs/docs/04-core-concepts/01-components.md
+++ b/docs/docs/04-core-concepts/01-components.md
@@ -84,7 +84,7 @@ func main() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/docs/04-core-concepts/01-components.md
```diff
@@ -82,6 +84,25 @@ func main() {
 }
 ```
 
+It is also possible to initlize a struct and call it's component method inline.
```
joerdav, line 87: "Initialize"

PR #568, opened 2024-02-28
fix: guard from index out of bounds error in function parsing

joerdav:
Sometimes a panic could surface with the following incomplete code:

```
package main

templ 
```


a-h (approved):
Nice!

Issue #492, opened 2024-02-04
generator: Support calls that return an error in attributes.


smit2909:
Sample code

```
import (
	"fmt"
)

func example() (string, error) {
	return "", fmt.Errorf("Not implemented yet.")
}

templ InvalidBlock() {
	<div boom={ example() }>SomeContent</div> //this complains,
}

templ ValidBlock() {
        <div>{ example() }</div> // this allows functions that return string, error
}
```

I see function expressions are supported in the actual value of the div tag, but attributes in a tag cannot support function expressions that return (string, error). I am wondering if this can be supported or is there a work around for this?

![image](https://github.com/a-h/templ/assets/79797111/25256fa6-e6dd-43b6-912a-7dea775259e2)

joerdav:
Hi @smit2909 , I think you may have encountered an oversight when introducing error handling to templates. Currently no, HTML attributes cannot include error results.

a-h:
Yes, I think it's a feature we'd want to add. Should be relatively straightforward to do, given the prior art of enabling it for `StringExpression`.

A change to the generator is required. For `StringExpression`, this is generated:

```go
var templ_7745c5c3_Var4 string
templ_7745c5c3_Var4, templ_7745c5c3_Err = templ.JoinStringErrs(doSomethingWithError("test"))
if templ_7745c5c3_Err != nil {
	return	templ.Error{Err: templ_7745c5c3_Err, FileName: ``, Line: 11, Col: 31}
}
```

For attribute expression, this is currently generated:

```go
_, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString(templ.EscapeString(doSomethingWithError("test")))
if templ_7745c5c3_Err != nil {
	return templ_7745c5c3_Err
}
```

The `JoinStringErrors` needs to be called, then the error check made, then output the results of the `templ.EscapeString` call.

smit2909:
Oh, thanks for clarifying, appreciate marking this as an enhancement. I wonder if this might also need changes to templ lsp.

a-h:
Because the templ LSP proxies go code through to gopls, it should be able to handle this without any changes.

sirgallifrey:
Took a shot at the fix #574

sirgallifrey:
@a-h this one can be closed due to #584

a-h:
Thanks for that!

PR #574, opened 2024-02-29
Fix: Support functions that return error in attributes

sirgallifrey:
My attempt at fixing #492

Tried to follow guidance on the original issue the best I could, let me know if I need to change something.
To ease the review, I created 2 commits, one with my fix and another with the new generated files.

Note: I didn't changed href attributes ad they require SafeUrl and also didn't changed script attributes.


generator/generator.go
```diff
@@ -1140,18 +1140,43 @@ func (g *generator) writeExpressionAttribute(indentLevel int, elementName string
...
+			if _, err = g.w.Write(")\n"); err != nil {
+				return err
+			}
+			// Attribute expression error handler.
+			_, err = g.w.WriteIndent(indentLevel, "if templ_7745c5c3_Err != nil {\n")
+			if err != nil {
+				return err
+			}
+			indentLevel++
+			_, err = g.w.WriteIndent(indentLevel, "return	templ.Error{Err: templ_7745c5c3_Err, FileName: "+createGoString(g.fileName)+", Line: "+strconv.Itoa(int(attr.Expression.Range.To.Line))+", Col: "+strconv.Itoa(int(attr.Expression.Range.To.Col))+"}\n")
```
sirgallifrey, line 1168: Note: I copied the code for getting the line number `attr.Expression.Range.To.Line` from the `writeStringExpression`, but I don't think this is correct, on my test it says the line is 17, but should actually be 18.

It is probably also wrong for `writeStringExpression`.
Do you folks want me to fix this? if yes, on this or another PR?

a-h: Great spot. It would be good to fix that. If you could fix that, I'd appreciate it!

Language Server Protocol starts line numbers from 0, but this is a runtime error where we expect an end user to read it, so they will expect lines to start from 1, like their text editor does.

generator/test-attribute-errors/render_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+		_, err := htmldiff.Diff(component, expected)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+	})
+	t.Run("attribute expressions can return errors", func(t *testing.T) {
+		errSomethingBad := errors.New("bad error")
+
+		err := render(errSomethingBad).Render(context.Background(), &bytes.Buffer{})
```
a-h, line 29: I think `render` is a bad name for a component. I think `component` would be better.

I probably used that name in the past, but now wish I hadn't. 😂

a-h (changes requested):
Great job. Sorry for the time to review, was prepping the last release.

If you've got time for the line number fix... brilliant. If not, let me know and I'll tackle it.

generator/test-attribute-errors/render_test.go
sirgallifrey: Yep, I copied that from some other test, but I can change, no problem

sirgallifrey: Changed to `TestComponent` as I saw in some other test files, let me know if you still you prefer `component`, I can change back

generator/generator.go
sirgallifrey: @a-h So, I fixed the line number, I don't know if I should do the same for the column.
Have a look as these two pictures: They are for the Attribute expression and for the String expressions, on both of them the line number is fixed and the column number is unchanged. The Line:Col as is now is ending right after `err` and before the `)`, not sure if that's the right place of if it should be after the `)`, but i think it should be after.
Screenshot 2024-03-03 at 20 10 26
Screenshot 2024-03-03 at 20 12 30

a-h (approved):

a-h:
Thanks a lot for this great contribution. I rebased it, and merged it in https://github.com/a-h/templ/pull/584

PR #583, opened 2024-03-03
docs: add more testing guidance

michaelByrne:
Integrated and adapted Adrian's [testing post](https://adrianhesketh.com/2021/10/17/testing-templ-html-rendering-with-goquery/) into the docs

### DIFF
--- a/docs/docs/04-core-concepts/03-testing.md
+++ b/docs/docs/04-core-concepts/03-testing.md
@@ -11,15 +11,17 @@ Expectation testing validates that the right data appears in the output in the r
 
 The example at https://github.com/a-h/templ/blob/main/examples/blog/posts_test.go shows how to test that a list of posts is rendered correctly.
 
-These tests use the goquery library to parse the HTML and check that the expected elements are present. goquery is a jQuery-like library for Go, and is useful for parsing and querying HTML. You’ll need to run `go get github.com/PuerkitoBio/goquery` to add it to your `go.mod` file.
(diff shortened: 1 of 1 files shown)


a-h:
Thanks for that!

michaelByrne:
> Thanks for that!

glad to help. any other issues or tasks you think would be good for me to work on? as somewhat of a newbie to the repo

a-h:
Thanks! If you're up for another docs change, yesterday, we added support for function calls in string attribute expressions, but there's no documentation for it at https://templ.guide/syntax-and-usage/attributes#string-expression-attributes

We try to keep the issues labelled appropriately - to identify stuff we need to think about, discuss, and label stuff that we think is OK to pick up.

However, I think we might do well to start running a regular planning session, like the Go team does with proposals, e.g. https://github.com/golang/go/issues/33502#issuecomment-1973634667

michaelByrne:
> Thanks! If you're up for another docs change, yesterday, we added support for function calls in string attribute expressions, but there's no documentation for it at https://templ.guide/syntax-and-usage/attributes#string-expression-attributes
> 
> We try to keep the issues labelled appropriately - to identify stuff we need to think about, discuss, and label stuff that we think is OK to pick up.
> 
> However, I think we might do well to start running a regular planning session, like the Go team does with proposals, e.g. [golang/go#33502 (comment)](https://github.com/golang/go/issues/33502#issuecomment-1973634667)

yeah, i can do that for sure! and i'd be down to join planning sessions as well

PR #595, opened 2024-03-06
feat: ComponentScript spread attribute support

alehechka:
# Overview
For context, there was a [thread ](https://gophers.slack.com/archives/C061C4PMKDJ/p1709649251693379) in the templ slack channel asking about support for `ComponentScript` attributes in the spread attribute functionality. This was just a quick exploratory attempt at this feature so I'm open to suggestions on the resulting generated code. 

The existing implementation only renders `<script>` elements from script components when directly provided in an `ExpressionAttribute` with the generated code as follows. 

```go
templ_7745c5c3_Err = templ.RenderScriptItems(ctx, templ_7745c5c3_Buffer, Alert("hello, world"))
	return templ_7745c5c3_Err
}
```

This proposed change will provide all spread attributes used within the element to a new `RenderScriptItemsWithSpread` that takes in a slice of `Attributes` and on runtime will loop over all attributes to discover any script components to render. 

```go
templ_7745c5c3_Err = templ.RenderScriptItemsWithSpread(ctx, templ_7745c5c3_Buffer, []templ.Attributes{attrs})
	return templ_7745c5c3_Err
}
```

Upon opening this PR, the implementation is set up so that providing spread attributes at all to an element will cause it to generate code with the new `RenderScriptItemsWithSpread` due to the unknown of what a parameter might pass in. 

> ⚠️ I still need to add unit tests over this new functionality but wanted to get the PR open early to get feedback on the implementation.


runtime.go
```diff
@@ -691,6 +695,20 @@ func (c ComponentScript) Render(ctx context.Context, w io.Writer) error {
 
 // RenderScriptItems renders a <script> element, if the script has not already been rendered.
 func RenderScriptItems(ctx context.Context, w io.Writer, scripts ...ComponentScript) (err error) {
+	return RenderScriptItemsWithSpread(ctx, w, nil, scripts...)
```
alehechka, line 698: I mainly only went this route to keep `RenderScriptItems` unchanged to avoid a breaking change. But if that's not a concern we can just add the extra parameter to this function.

Issue #597, opened 2024-03-06
0.2.598 parser is slower than 0.2.543 on large files


dv1x3r:
Hello, I noticed that the template generation works slower with version 598 on large .templ files.

I have a templ file with ~1k rows, and ~30 components. `templ generate` is ~5x slower on this one.

Here is the comparison before the update:

```bash
╰─ templ version
v0.2.543

╰─ time templ generate
Processing path: /Users/dx/source/teh-go
Generating production code: /Users/dx/source/teh-go
(✓) Generated code for "/Users/dx/source/teh-go/internal/app/component/admin.templ" in 2.520125ms
(✓) Generated code for "/Users/dx/source/teh-go/internal/app/component/client.templ" in 46.07ms
(✓) Generated code for 2 templates with 0 errors in 75.794833ms
templ generate  0.09s user 0.06s system 163% cpu 0.091 total
```

And after the update:

```bash
╰─ templ version
v0.2.598

╰─ time templ generate -v
(✓) Starting post-generation handler
(✓) Walking directory [ path=/Users/dx/source/teh-go devMode=false ]
(✓) Starting event handler
(✓) Processing file [ file=/Users/dx/source/teh-go/internal/app/component/admin_templ.go ]
(✓) Processing file [ file=/Users/dx/source/teh-go/internal/app/component/client_templ.go ]
(✓) Processing file [ file=/Users/dx/source/teh-go/internal/app/component/client.templ ]
(✓) Processing file [ file=/Users/dx/source/teh-go/internal/app/component/admin.templ ]
(✓) Generated code [ file=/Users/dx/source/teh-go/internal/app/component/admin.templ in=2.148833ms ]
(✓) First post-generation event received, starting proxy
(✓) No proxy URL specified, not starting proxy
(✓) Dev mode not enabled, process can finish early
(✓) Generated code [ file=/Users/dx/source/teh-go/internal/app/component/client.templ in=4.215424s ]
(✓) Post-generation event channel closed, exiting
(✓) Waiting for push handler to complete
(✓) Waiting for event handler to complete
(✓) Waiting for post-generation handler to complete
(✓) Complete [ updates=2 duration=4.216013041s ]
templ generate -v  5.84s user 0.95s system 160% cpu 4.231 total
```

I can split the `client.templ` into several smaller pieces, and then it will become "acceptable".
But I do not want to do that for this specific project.

```bash
╰─ templ version
v0.2.598

╰─ time templ generate -v
(✓) Starting post-generation handler
[...]

PR #605, opened 2024-03-10
fix: fixes issue #597 - improves performance (500x faster than previous)

a-h:


joerdav:
All the code looks solid to me, I also tested it on my shopping-list repo, which doesn't really benefit form the performance increase much, but it does still work.

One thing with these open/close counters is I get skittish about some scenarios like `f("hello world{")`, where strings contain open or close brackets. Looking at the code I think we're fine, but could do with a few test cases to be positive.

Would also probably be an ideal time to run the fuzz tests again!

joerdav (approved):

PR #615, opened 2024-03-14
feat: add `--proxybind`, giving the option to bind to something else than `127.0.0.1`

AlyxPink:
I love this proxy option, it really makes web development easier! So first, thank you for creating `templ` and adding this awesome feature 🙇 

# Overview

I wanted to use it with docker-compose, but because it's binding to `127.0.0.1` only, I could not use it from another container.

# Use case
Here's a docker-compose.yaml example to show how it would be used:

```yaml
version: "3.8"
services:
  web:
    volumes:
      - ./:/src/
    ports:
      - "3000:3000"
    command: "go run /src/cmd/web/"

  templ:
    volumes:
      - ./:/src/
    ports:
      - "7331:7331"
    command: 'templ generate --watch --proxy="http://web:3000" --proxybind="0.0.0.0" --cmd="go run ./cmd/web/"'
```

Let me know what you think, if I can add any tests, etc 😊

### DIFF
--- a/cmd/templ/generatecmd/cmd.go
+++ b/cmd/templ/generatecmd/cmd.go
@@ -301,10 +301,13 @@ func (cmd *Generate) StartProxy(ctx context.Context) (p *proxy.Handler, err erro
 	if cmd.Args.ProxyPort == 0 {
 		cmd.Args.ProxyPort = 7331
 	}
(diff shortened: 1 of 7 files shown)


a-h (approved):
Nice improvement. Thanks a lot, and thanks for your nice comments too. 😁

PR #648, opened 2024-03-28
Handle string and boolean pointers in attributes

### LINKED ISSUE #647: generator: *string attribute value is ignored
Right now `templ.Attributes` silently ignores `*string` during rendering. It would be helpful if it had a case for `*string` and if the value is not nil render it

whalenrp:
I took a stab at adding support for this, but let me know if this isn't the behavior we'd expect from the API. 

Resolves #647


a-h:
LGTM, thanks for that.

PR #657, opened 2024-03-30
Allow using `tel`, `ftp` and `ftps` as safe URL protocols

### LINKED ISSUE #655: refactor: add `tel:` as safe URL protocol
templ: v0.2.648

Currently trying to use this:
`templ.URL("tel:" + phoneNumber)`

results in:
`about:invalid#TemplFailedSanitizationURL`

I think this is because it only allows `http`, `https` and `mailto`, but I think `tel` should be added to this list. It's like `mailto` but for telephone numbers, so it will open up a phone application.

atifcppprogrammer:
Closes #655


joerdav (approved):

a-h:
After spotting this, I added a unit test and benchmark this morning. I was interested in whether it was a good idea to just add another bunch of if checks, rather than look up in a map or slice, but it was still faster to keep the if statements.

PR #661, opened 2024-04-02
[feat] implement notify proxy to reload browser

### LINKED ISSUE #656: cmd: add `--notify` to `generate --watch` in cli
## The problem

Templ handles automatic reloading of browser based on SSE when it detects a file change. However, it only watches Templ related files, and rightly so. The community has come up with various methods to watch other files e.g. for Go, there is `air` and `wgo`, for tailwindcss, there is tailwindcss cli. For example, this is my solution as an example:

```makefile
live/templ:
	templ generate --watch --proxy="http://localhost:8080" --open-browser=false -v

live/server:
	go run github.com/cosmtrek/air@v1.43.0 \
	--build.cmd "make build" --build.bin "/tmp/bin/${BINARY_NAME}" --build.delay "100" \
	--build.exclude_dir "assets,node_modules,bak" \
	--build.include_ext "go,templ" \
	--misc.clean_on_exit "true"

live/tailwind:
	npx tailwindcss -i ./input.css -o ./assets/styles.css --minify --watch

live: 
	make -j3 live/templ live/server live/tailwind
```

Here I start 3 watch processes:

1. templ generate
2. air for rebuilding and reloading go
3. tailwindcss watch

However, the reloading of the browser doesn't always work. It sort of works for Go files, take this for an example:

1. edit and save index.templ
2. templ watches and re-generates index_templ.txt. Around the same time, air detects templ change and tries to rebuild and reruns the binary 
3. templ send down reload event to the browser via SSE

In this process, go build and restart proces is usually very fast, so the reload usually happens correctly. But for reloading the `assets/styles.css` when tailwind rebuilds it, there is no right way to do it.

Templ would like to stay simple to only handle its related files, see issue https://github.com/a-h/templ/issues/646 for which I fully support. However templ should provide a way for other process to notify there is a change so that users can progmatically send the reload event when necessary.

## Proposal

Add `--notify` to `generate` subcommand. An example would look like this:

```sh
templ generate --watch --notify
```

It should send an reload event via SSE and exit immediately. By default, it sends to port 7331, for an alternative port, `-proxyport` can be specified.

If this is implemented, one can simply add another directive in the makefile like so:

```makefile
live/reload:
	wgo -dir assets temple generate --watch --notify

live: 
	make -j4 live/templ live/server live/tailwind live/reload
```

## Alternatives

Alternative I considered is [browser-sync](https://github.com/Browsersync/browser-sync), however it can't handle templ related generation efficiently.

jackielii:
fix #656

### DIFF
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -272,7 +272,7 @@ func TestProxy(t *testing.T) {
 				}
 			}
 			err = scanner.Err()
-			// we expect the connection to be closed by the server: this is the only way to terminate the sse connection
+			// We expect the connection to be closed by the server: this is the only way to terminate the sse connection.
 			if err != nil {
 				errChan <- err
 				return
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -240,7 +240,7 @@ func TestProxy(t *testing.T) {
 			t.Fatalf("unexpected error parsing port: %v", err)
 		}
 
-		// Arrange 2: start a goroutine to listen for sse events
+		// Arrange 2: start a goroutine to listen for sse events.
 		ctx := context.Background()
 		ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 		defer cancel()
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -254,7 +254,7 @@ func TestProxy(t *testing.T) {
 				errChan <- err
 				return
 			}
-			resp, err := http.DefaultClient.Do(req.WithContext(ctx))
+			resp, err := http.DefaultClient.Do(req)
 			if err != nil {
 				errChan <- err
 				return
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -280,7 +280,7 @@ func TestProxy(t *testing.T) {
 		}()
 
 		// Act: notify the proxy.
-		select { // either sse is listening or an error occurred
+		select { // Either SSE is listening or an error occurred.
 		case <-sseListening:
 			err = NotifyProxy(u2.Hostname(), port)
 			if err != nil {
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -218,6 +218,7 @@ func TestProxy(t *testing.T) {
 	})
 
 	t.Run("notify-proxy: sending POST request to /_templ/reload/events should receive reload sse event", func(t *testing.T) {
+		// Arrange 1: create a test proxy server
 		dummyHandler := func(w http.ResponseWriter, r *http.Request) {}
 		dummyServer := httptest.NewServer(http.HandlerFunc(dummyHandler))
 		defer dummyServer.Close()
@@ -239,7 +240,7 @@ func TestProxy(t *testing.T) {
 			t.Fatalf("unexpected error parsing port: %v", err)
 		}
 
-		// we start an sse request to receive the events. The connection will be closed by the server or timeout.
+		// Arrange 2: start a goroutine to listen for sse events
 		ctx := context.Background()
 		ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
 		defer cancel()
@@ -278,8 +279,8 @@ func TestProxy(t *testing.T) {
 			}
 		}()
 
-		// either sse is listening or an error occurred
-		select {
+		// Act: notify the proxy
+		select { // either sse is listening or an error occurred
 		case <-sseListening:
 			err = NotifyProxy(u2.Hostname(), port)
 			if err != nil {
@@ -291,8 +292,8 @@ func TestProxy(t *testing.T) {
 			}
 		}
 
-		// either sse has a expected response or an error or timeout occurred
-		select {
+		// Assert
+		select { // either sse has a expected response or an error or timeout occurred
 		case resp := <-sseRespCh:
 			if !strings.Contains(resp, "event: message\ndata: reload") {
 				t.Errorf("expected sse reload event to be received, got: %q", resp)
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -1,13 +1,19 @@
 package proxy
 
 import (
+	"bufio"
 	"bytes"
 	"compress/gzip"
+	"context"
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 )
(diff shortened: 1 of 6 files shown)


jackielii:
Let me know if anything is needed. I plan add some example setups for reloading script in typical set. E.g. tailwindcss + npm build + templ + go live reload example. But I think that's better suited in the docs using another PR

a-h:
It looks great! Thanks a lot for this.

I'd love to see an integration test where it starts a web server to listen to the POST request, runs the CLI `go run cmd/templ --notify-proxy`, and validates that the post was received.

If you haven't got time, I can do that though, just let me know!

jackielii:
> It looks great! Thanks a lot for this.
> 
> I'd love to see an integration test where it starts a web server to listen to the POST request, runs the CLI `go run cmd/templ --notify-proxy`, and validates that the post was received.
> 
> If you haven't got time, I can do that though, just let me know!

I can write a http recorder test to test the new route handle with the notify function call. Would that be enough? Probably much easier to write?

a-h:
Great idea. The main reason is to protect against future regressions.

jackielii:
added a httptest.Server test. Hopefully I didn't miss anything. I find it wired that URL appears in the handler: https://github.com/a-h/templ/blob/main/cmd/templ/generatecmd/proxy/proxy.go#L29. I know it's being used inside the `proxy.Handler` attached methods, but handler doesn't really care where the request come from right?

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
 			t.Errorf("unexpected response body (-got +want):\n%s", diff)
 		}
 	})
+
+	t.Run("notify-proxy: sending POST request to /_templ/reload/events should receive reload sse event", func(t *testing.T) {
+		// Arrange 1: create a test proxy server
```
a-h, line 221:
```suggestion
		// Arrange 1: create a test proxy server.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+		u2, err := url.Parse(proxyServer.URL)
+		if err != nil {
+			t.Fatalf("unexpected error parsing URL: %v", err)
+		}
+		port, err := strconv.Atoi(u2.Port())
+		if err != nil {
+			t.Fatalf("unexpected error parsing port: %v", err)
+		}
+
+		// Arrange 2: start a goroutine to listen for sse events
```
a-h, line 243:
```suggestion
		// Arrange 2: start a goroutine to listen for sse events.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("unexpected error notifying proxy: %v", err)
+			}
+		case err := <-errChan:
+			if err == nil {
+				t.Fatalf("unexpected sse response: %v", err)
+			}
+		}
+
+		// Assert
```
a-h, line 295:
```suggestion
		// Assert.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+		}
+
+		// Arrange 2: start a goroutine to listen for sse events
+		ctx := context.Background()
+		ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+		defer cancel()
+
+		errChan := make(chan error)
+		sseRespCh := make(chan string)
+		sseListening := make(chan bool) // coordination channel to ensure the sse listener is started before notifying the proxy
```
a-h, line 250:
```suggestion
		sseListening := make(chan bool) // Coordination channel that ensures the SSE listener is started before notifying the proxy.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+			scanner := bufio.NewScanner(resp.Body)
+			for scanner.Scan() {
+				lines = append(lines, scanner.Text())
+				if scanner.Text() == "data: reload" {
+					sseRespCh <- strings.Join(lines, "\n")
+					return
+				}
+			}
+			err = scanner.Err()
+			// we expect the connection to be closed by the server: this is the only way to terminate the sse connection
```
a-h, line 275:
```suggestion
			// We expect the connection to be closed by the server: this is the only way to terminate the sse connection.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+			if err != nil {
+				errChan <- err
+				return
+			}
+		}()
+
+		// Act: notify the proxy
```
a-h, line 282:
```suggestion
		// Act: notify the proxy.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+		select { // either sse is listening or an error occurred
```
a-h, line 283:
```suggestion
		select { // Either SSE is listening or an error occurred.
```

cmd/templ/generatecmd/proxy/proxy_test.go
```diff
@@ -210,4 +216,94 @@ func TestProxy(t *testing.T) {
...
+				t.Fatalf("unexpected error notifying proxy: %v", err)
+			}
+		case err := <-errChan:
+			if err == nil {
+				t.Fatalf("unexpected sse response: %v", err)
+			}
+		}
+
+		// Assert
+		select { // either sse has a expected response or an error or timeout occurred
```
a-h, line 296:
```suggestion
		select { // Either SSE has a expected response or an error or timeout occurred.
```

a-h (approved):

docs/docs/09-commands-and-tools/03-hot-reload.md
leandergangso, line 1: Great work, this is a nice addition to the templ CLI, will be using this once it's released.
Maybe even replace `--watch`?

I was thinking that a change to the explanation, to make it more clear, could help this feature be more utilized when it's released.

By specifying that `--notify-proxy` takes priority over other flags, and will be exec just once, can help developers better understand the usage for this cmd.
Also, the first time I read this text, it felt like I should add `--notify-proxy` as a flag to the `templ generate --watch` cmd, which would end in an unexpected outcome. (perhaps just my thinking tho :P )

Could also mention that `--notify-proxy` only works if `--proxy` is used. You kinda already does this and it's somewhat self explanatory, but one extra sentence could clear up any misunderstandings?

PS: some of this can maybe also be applied to the `--notify-proxy` CLI help text.

This was just some thoughts on my part.

docs/docs/09-commands-and-tools/03-hot-reload.md
```diff
@@ -82,6 +82,20 @@ sequenceDiagram
...
+```shell
+templ generate --notify-proxy
+```
+
+This will default to the default templ proxy address of `localhost:7331`, but can be changed with the `--proxybind` and `--proxyport` arguments.
+
+```shell
+templ generate --notify-proxy --proxybind="localhost" --proxyport="8080"
+```
+
```
leandergangso, line 98: Great work, this is a nice addition to the templ CLI, will be using this once it's released.
Maybe even replace --watch?

I was thinking that a change to the explanation, to make it more clear, could help this feature be more utilized when it's released.

By specifying that --notify-proxy takes priority over other flags, and will be exec just once, can help developers better understand the usage for this cmd.
Also, the first time I read this text, it felt like I should add --notify-proxy as a flag to the templ generate --watch cmd, which would end in an unexpected outcome. (perhaps just my thinking tho :P )

Could also mention that --notify-proxy only works if --proxy is used. You kinda already does this and it's somewhat self explanatory, but one extra sentence could clear up any misunderstandings?

PS: some of this can maybe also be applied to the --notify-proxy CLI help text.

This was just some thoughts on my part.

PR #666, opened 2024-04-04
docs: Create documentation and examples for Internationalization

### LINKED ISSUE #154: docs: create an example integration with I18n for translations
Wondering if language translation workflow has been thought about yet ? 

Maybe a proper way to do it leveraging the architecture of templ.

I normally Setup these systems to do real time machine translation and then a person can check it later.

vigneshsankariyer1234567890:
Our current version of templ does not feature examples for internationalization, which could help with adoption.

This PR aims to resolve that by
1) Adding an example for i18n using [ctxi18n](https://github.com/invopop/ctxi18n)
2) Adding documentation under Integrations to use [ctxi18n](https://github.com/invopop/ctxi18n)

For the example: I created an SSR app with a title and a few buttons. Languages could be switched based on provided path parameters (e.g. /en gives English translations, whereas /de gives German translations etc.)

For the documentation: I added the internationalization documentation as a subpage under Integrations, while renaming the existing page as "Web frameworks" 

Closes #154


a-h:
Thanks for that!

PR #668, opened 2024-04-04
remove redudent comment

jackielii:
@a-h Thanks for merging #661. I realised I left a redundant comment here.

I used testServer to terminate all connection after I notify proxy, but realised it doesn't work reliably. So now the test sse request listening goroutine will wait for expected result, or timeout to close the connection from the client side. Then the channels will receive and asset and close the test server in defer.


a-h (approved):

a-h:
Thank you for the contribution, I really appreciate it!

PR #670, opened 2024-04-05
fix: allow slice type definitions in templ expressions

### LINKED ISSUE #669: parser: Issue parsing slice types in templ expressions
Hi,

First - I couldn't think of a better title for the issue since I'm not yet sure what's happening. I'm sorry about that.

The latest version (`v0.2.648`) of templ seems to be having issues with slice types / slices in some cases. I've tried to produce a small example of the issue:

```go
package example

import (
	"fmt"
)

templ componentA(x func()) {
}

templ Example() {
	@componentA(func() {
		fmt.Println(make([]int))
	})
	
}
```

If this is placed in an `example.templ` file `templ generate` produces the following error:

```
example/example.templ source formatting error: 51:1: expected statement, found '.'
```

If I replace the `[]int` type with something else that isn't a slice type (i.e., `map[string]string`) it works fine.

Replacing the `make` call with something else, like a string, works fine as well.

I've also opened a thread for this on Slack where @joerdav stated https://github.com/a-h/templ/blob/main/parser/v2/goexpression/parse_test.go#L474 as a related resource (parsing tests, which are probably missing cases related to this).

Go version: `go version go1.22.1 darwin/arm64`
templ version: `v0.2.648`

joerdav:
Fixes #669

### DIFF
--- a/parser/v2/goexpression/parse_test.go
+++ b/parser/v2/goexpression/parse_test.go
@@ -454,6 +454,12 @@ var templExpressionTests = []testInput{
 })`,
 	},
 	{
+		name: "function call with function arg",
+		input: `componentA(func(y []int) string {
+		return "hi"
+	})`,
+	},
+	{
 		name:  "call with braces and brackets",
 		input: `templates.New(test{}, other())`,
 	},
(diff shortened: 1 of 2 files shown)


joerdav:
@a-h I couldn't think of any cases that might break this, can you?

a-h:
No, I just wanted to get another release out before merging this one, so that it has time to be tested alongside other new changes.

PR #687, opened 2024-04-17
fix: send SIGTERM signal to --cmd instead of SIGKILL

guilhas07:
### Problem 
When trying to use the command `templ generate --watch --proxy="http://localhost:8000" --cmd="wgo run main.go"` I noticed that wgo didn't close correctly. After discussing on the wgo repo ( see [here](https://github.com/bokwoon95/wgo/issues/8) ) the conclusion was that this was due to templ sending a SIGKILL signal not allowing for wgo to close correctly.

### Solution
To fix this I changed all SIGKILL signals to SIGTERM. 

Caveats:
- I don't know if this is a problem too on windows
- I suppose this doesn't break any behaviour, but let me know if it does. 

Let me know if there is something I need to change!

### DIFF
--- a/.direnv/flake-profile
+++ b/.direnv/flake-profile
@@ -1,0 +1,1 @@
+flake-profile-12-link
--- a/.direnv/flake-profile
+++ b/.direnv/flake-profile
@@ -1,1 +1,0 @@
-flake-profile-12-link
--- a/.direnv/flake-profile-12-link
+++ b/.direnv/flake-profile-12-link
@@ -1,0 +1,1 @@
+/nix/store/div4cqrcwby72ckvcxyw317i9vzlm3n5-nix-shell-env
--- a/.direnv/flake-profile-12-link
+++ b/.direnv/flake-profile-12-link
@@ -1,1 +1,0 @@
-/nix/store/div4cqrcwby72ckvcxyw317i9vzlm3n5-nix-shell-env
--- a/cmd/templ/generatecmd/run/run_test.go
+++ b/cmd/templ/generatecmd/run/run_test.go
@@ -1,0 +1,108 @@
+package run_test
+
+import (
+	"context"
+	"embed"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/a-h/templ/cmd/templ/generatecmd/run"
+)
+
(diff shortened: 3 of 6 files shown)


a-h:
Thanks, I'll have a read through of the other repo. Sigkill kills child processes, where sigterm doesn't, so there's a good reason to use kill - for example, if you do `go run`, it does a build, then runs the executable as a child process, so sigterm wouldn't necessarily kill that process.

guilhas07:
> Thanks, I'll have a read through of the other repo. Sigkill kills child processes, where sigterm doesn't, so there's a good reason to use kill - for example, if you do `go run`, it does a build, then runs the executable as a child process, so sigterm wouldn't necessarily kill that process.

Using `go run` seems to be working as expected with this version too. Using this [example](https://github.com/bokwoon95/wgo/issues/8#issuecomment-2058569487) mentioned on the issue above.

a-h:
Reading through Air's code, it appears that it can send a SIGTERM to initiate graceful shutdown, prior to attempting to forcefully kill the process after a timeout: https://github.com/cosmtrek/air/blob/3f19370fe5e8d8fe2ddd927de54b1aede379f2b7/runner/util_unix.go#L13

I think that's what templ needs to do in order to allow wgo to shutdown any processes it's started gracefully.

guilhas07:
> Reading through Air's code, it appears that it can send a SIGTERM to initiate graceful shutdown, prior to attempting to forcefully kill the process after a timeout: https://github.com/cosmtrek/air/blob/3f19370fe5e8d8fe2ddd927de54b1aede379f2b7/runner/util_unix.go#L13
> 
> I think that's what templ needs to do in order to allow wgo to shutdown any processes it's started gracefully.

I believe the current way in this PR should be correct because we should not assume how long the user command takes in order to execute its clean-up logic. ref: https://stackoverflow.com/a/690631/12458551

a-h:
Hi, it took me a while to get to this. I've added a test which I think demonstrates the behaviour we're looking for.

Is this OK for you?

guilhas07:
> Hi, it took me a while to get to this. I've added a test which I think demonstrates the behaviour we're looking for.
> 
> Is this OK for you?

No worries! Working well on my end. 
Thank you for looking into this!

joerdav (approved):
Looks good to me!

PR #688, opened 2024-04-17
feat: support build tags by copying template header over

### LINKED ISSUE #625: feature: go:build tags support?
go has these handy `//go:build <tag>` comments, allowing stuff like this:

`config_dev.go`
```go
//go:build  dev

package config

const Port = 8080
```

`config.go`
```go
//go:build  !dev

package config

const Port = 80
```

sadly, templ doesnt seem to support that. 
right now i'm using a bool variable "AppEnvDev" in a config and do stuff like this:

 `page.templ`
```go
package views

...

templ js() {
	if config.AppEnvDev {
		<script src="main.js"></script>
	} else {
		<script src="main.min.js"></script>
	}
}
```

does it job, but dev stuff still gets shipped in the final binary.

any possbility to support said tags? i'm not familiar how go build tags work with generated code, so dont know about any limitation

joerdav:
Closes #625 

Excuse the diff, my formatter got a bit trigger happy. I also moved some tests to the txtar format.


a-h:
Would you be able to stop that formatter from putting args on new lines? It looks like it's optimising for 80 chars. 😁

joerdav:
Sure, I always use `gofumpt` which is quite opinionated.

joerdav:
Done!

a-h (approved):

PR #691, opened 2024-04-21
Fix link for "CSS style management"

AlyxPink:
Found a 404 on https://templ.guide/syntax-and-usage/attributes/ so here's the fix 🙌


a-h:
Thanks!

PR #695, opened 2024-04-22
feat: add brotli support to proxy, warn on unsupported encoding

### LINKED ISSUE #692: Proxy: injection of `script.js` with Brotli compression is silently failing
I encountered an issue where `script.js` was inconsistently injected into pages across different browsers. Specifically, both Safari and cURL successfully had the JavaScript script injected, whereas Firefox did not.

Upon further investigation by exporting requests from Safari and Firefox as cURL commands, I pinpointed the cause to the differences in the `Accept-Encoding` header. The request from Firefox uses `Accept-Encoding: br`, unlike Safari or a basic cURL command.

Here are the results demonstrating this discrepancy:
```
$ curl -s 'http://localhost:7331/' -H 'Accept-Encoding: gzip' | gunzip - | grep -Eo '</main>.*'
</main><script src="/_templ/reload/script.js"></script></body></html>
```

```
$ curl -s 'http://localhost:7331/' -H 'Accept-Encoding: br' | brotli -d | grep -Eo '</main>.*'
</main></body></html>
```

I don't think I would be capable of submitting a PR, but I'll give a try if I got time. Maybe warning the user that only `gzip` compression is supported today could be easily doable, while a proper fix could come later?

a-h:
- feat(proxy): add brotli support, fixes #692
- feat(proxy): warn on unsupported encoding, fixes #692


joerdav (approved):

AlyxPink:
Just for reference, I think there is a last issue to fix: https://github.com/a-h/templ/issues/692#issuecomment-2078474730

PR #696, opened 2024-04-22
feat: normalise line breaks and path separators to Linux style, on Windows, fixes #539

### LINKED ISSUE #539: Generate output differs between windows and linux
Seems mostly related to line endings and path separators in errors. I understand that this behavior could technically be considered correct but personally I think deterministic generation from a given input file would be preferred.

image
image

Both of these screenshots came from autogen diff with no source diff.

a-h:


joerdav (approved):

joerdav:
Nice!

PR #697, opened 2024-04-22
refactor: remove legacy v1 to v2 migration feature

a-h:


joerdav:
Do you think its worth a README or docs entry? Just to say "if you're using v1 its no longer supported, install version x for the migration tool"

Or maybe we can create a separate repo to preserve the migration tool but make it clear it will only migrate so far and won't be updated?

a-h:
I figured that if you were still on v1 and wanted to migrate, you could use an old version of templ to do it. Worth a docs update, yes, I'll do that.

joerdav (approved):

PR #701, opened 2024-04-25
docs: Expand Syntax & Usage > Template Composition to show how to import components from different packages

gabrielvincent:
This PR adds documentation to explicitly show how to import Templ components from other packages in the same project. The idea for this PR came up in this discussion: https://github.com/a-h/templ/discussions/554#discussioncomment-9178708.


a-h (approved):
I've basically rewritten it, but your PR triggered me to think about it more!

a-h:
Thanks!

PR #713, opened 2024-04-30
feat: add arbitrary go code support

### LINKED ISSUE #109: generator: allow go code statements in templates
In quick template you can add go vote anywhere in template using current scope.  Doesn't appear to be possible here.  Am I missing something?

### LINKED ISSUE #399: generator: allow arbitrary go code in templates
It would be awesome if you could define a temporary variable somehow (this is just one possible suggestion) in cases where you might be duplicating lots of code to reuse something, like this:

```go
package views

templ ItemDetail(item Item) {
	t := item.CreatedAt.Local()
	<section>
		<h1>{ item.Name }</h1>
		<time datetime={ t.Format(time.RFC3339) }>{ t.Format("Jan 2, 2006") }</time>
	</section>
}
```

Unless I'm mistaken, this doesn't appear to be supported - so if I'm correct - I'd love to see this implemented some day!

joerdav:
Closes #109  

TODO:

- [ ] Add docs
- [x] Add feature flag


joerdav:
I've made a first pass at the experiment functionality. Next I'll look at documentation.

I do think that package level vars are the best way to do this without affecting the rest of the code massively, or causing a performance impact. Let me know what you think, happy to change, this approach was made basically copying and simplifying the GOEXPERIMENT code.

a-h:
I tried to add a generator test, but realised that the generator test couldn't run without the experiment being enabled, which complicated execution of the CI/CD processes, so best leave it for now.

I renamed `TEMPLEXPERIMENT` because it was messing with my eyes. I kept seeing TEMPLEX.

But I noticed the formatting of `b := func()` got a little messed up by the formatter.

```templ
package testfor

templ render(items []string) {
	{{ a := "a" }}
	<div>{ a }</div>
	{{b := func() string {
	return "b"
}
	}}
	<div>{ b() }</div>
	{{
	c := func() (string, error) {
		return "c", nil
	}
	}}
	<div>{ c() }</div>
}
```

Looks great!

a-h (approved):

joerdav:
Good shout on those amends! Maybe worth us adding an experiment field to the editor plugins that allow for config so things like this can easily be enabled on the LSP.

adpande:
This is really awesome! I am already using it to generate random strings as ids in components. I dont have to pass them now as function parameters.

PR #714, opened 2024-04-30
feat: templ component snippet

valentimarco:
Refer to this [PR](https://github.com/templ-go/templ-vscode/pull/54) for more info.

P.S i don't understand how to test this piece of code... I was able to build the command but idk how to connect vscode (in my case) to the lsp server


joerdav:
I don't think that `package ${1:};` should be included. I think that the templ snippet should only insert one templ element.

And in terms of connecting to vscode, do you get autocomplete in templ files?

valentimarco:
> I don't think that `package ${1:};` should be included. I think that the templ snippet should only insert one templ element.
> 
> And in terms of connecting to vscode, do you get autocomplete in templ files?

Ok make sense to delete the declaration of the package! For vscode, i dont get the autocomplete bc idk how to connect the lsp server to vscode... I tried with the plugin, but the http param in the settings open the lsp server, instead i want to connect to an existing lsp server

joerdav:
The extension should start the server for you, sometimes it helps to start vscode from the command line so that it has templ in the path `code .`

joerdav:
I also realize I sent you to the wrong part of the code base, this is for html snippets only (sorry). 

In this bit of code https://github.com/a-h/templ/blob/main/cmd/templ/lspcmd/proxy/server.go#L361 it returns completion items, essentially we want to append the completion item to the result of this function.

valentimarco:
> I also realize I sent you to the wrong part of the code base, this is for html snippets only (sorry).
> 
> In this bit of code https://github.com/a-h/templ/blob/main/cmd/templ/lspcmd/proxy/server.go#L361 it returns completion items, essentially we want to append the completion item to the result of this function.

Ok i figure the code and i added this line after the for line 388:
```golang

result.Items = append(result.Items, lsp.CompletionItem{
		Label: "templ",
		InsertText: `templ ${2:}() {
	${0}
}`,
		Kind:             lsp.CompletionItemKind(lsp.CompletionItemKindSnippet),
		InsertTextFormat: lsp.InsertTextFormatSnippet,
	})
```
And works! 
But i have a bug where the `templ` is not highlight until i write the template name (vscode snippet don't create this problem)

joerdav:
I'm happy with this, I think its a useful snippet to have. What to you think @a-h ?

joerdav:
One of the tests now fails, I think the expectation needs updating.

joerdav:
Yeah, so there is a test that is expecting no snippets to be returned, but now we have a global snippet that is returned every time.

I'd recommend renaming the test since the name is no longer true. And updating the assertion to expect your new snippet.

You can run the tests with `go test ./...`

joerdav:
Let me know if you need a hand, then I can fix it up.

valentimarco:
I got so much confused when trying to create the test, so far:
1. Deleted the test that create the problem
2. trying to copy the test logic without success

joerdav:
Thanks for this!

cmd/templ/lspcmd/lsp_test.go
```diff
@@ -100,12 +99,15 @@ func TestCompletion(t *testing.T) {
...
-			replacement: ` <div data-testid="count">{  `,
-			cursor:      `                            ^`,
+			line:        3,
+			replacement: `templ`,
+			cursor:      `    ^`,
 			assert: func(t *testing.T, actual *protocol.CompletionList) (msg string, ok bool) {
-				if diff := lspdiff.CompletionList(nil, actual); diff != "" {
-					return fmt.Sprintf("unexpected completion: %v", diff), false
+				if !lspdiff.CompletionListContainsText(actual, "templ") {
+					return fmt.Sprintf(
```
a-h: Would really prefer that this formatting wasn't applied.

joerdav: I'll figure out a whitelist, I keep forgetting to toggle it.

joerdav: Ah, sorted, it wasn't gofumpt after all, I had the golines formatter enabled too.

PR #723, opened 2024-05-03
feat(parser): support colons in element names

deej-io:
I am using templ to generate HXML for a [Hyperview](https://hyperview.org/) app I am building (blog post about it coming soon :crossed_fingers:).

[HXML](https://hyperview.org/docs/reference_custom_elements#using-in-hxml) supports custom elements via a namespace - represented by a colon in the tag name.

I couldn't see anything in the HTML5 spec about whether this is valid, and I was worried how it would work with pseudo elements, but it does seem to be handled by browsers as per this example (tested in both Firefox and Chromium):

```html
 <!DOCTYPE html>
<html>
<head>
	 <style>
		foo\:bar {color: red;}
		foo\:bar:hover {color: blue;}
	</style>
</head>
<body>

<div>
	<foo:bar>
		<h1>Hello</h2>
	</foo:bar>
</div>

</body>
</html>
```


joerdav:
I think we're correctly not parsing if the name starts with a colon, but it seems also not valid for it to end with a colon. But could be argued its not something templ should enforce?

a-h (approved):

a-h (approved):

deej-io:
Thanks for merging, folks!

joerdav:
Looking forward to the blog post :)

PR #725, opened 2024-05-03
fix: go escape constant attributes

### LINKED ISSUE #624: bug: backslashes in constant string attributes cannot be generated
This issue is circling back to to https://github.com/a-h/templ/issues/498

On top of Lsp support, HTML's input element has a built in attribute [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) which:
- allows us to validate user's input in the browser with regular expression
- custom error message on input field with [setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity)

Previously I had a case of validating url links and this pattern `(https?|ftp):\/\/[^\s/$.?#].[^\s]*` worked in javascript, but it doesn't work in templ or Golang (tried passing it from backend as a string). templ lsp and parser will warn me with a red flag and causing generation failure at `\s`. 

The alternative of not having this feature is either use additional javascript + bundle to add input validation or using [validation from htmx](https://htmx.org/docs/#validation) where the lsp + syntax highlighting features are not yet complete.

Robert-Kolmos:
Fixes #624, Go escapes constant attribute strings.

This changes the previous behavior where these strings were treated as go strings. If someone previously had

```
templ foo() {
    <input pattern="\\s">
}
```
this will now generate 
`<input pattern="\\s"/>` instead of `<input pattern="\s"/>` like the current latest version would generate.


a-h (approved):

a-h:
Thanks a lot for the contribution. Much appreciated!

PR #729, opened 2024-05-06
fix: set headers before calling WriteHeader

bradydean:
While WriteHeader correctly infers the Content-Type as HTML in the example, the correct way to use ResponseWriter is to set the response headers first, then call WriteHeader.


a-h:
Thanks for the contribution! 😁

PR #734, opened 2024-05-12
chore: update echo example

bradydean:
The previous implementation did not play nice with echo's error handler. The error handler does not override an already-set content-type, nor the response code (if already written to the underlying ResponseWriter), so error responses looked really weird. Eg you could get a code 200, content-type: text/html, with a json body.

This PR scraps that, and uses an implemtation really similar to what `templ.Handler` does, which is to allocate a byte buffer and render the component to that. This plays nice with the error handler and doesn't return any nonsense.


a-h:
Thanks!

Issue #726, opened 2024-05-04
Errors in go portions of a template report line numbers in the go file instead of the templ file


Robert-Kolmos:
Hey all, 

I've got a very simple templ file

```
package main

// : after parameter name is intentional
templ Template(invalid: string) {
	<div/>
}
```

When I run `templ generate` I get an error like the following

```
(✗) Error generating code [ file=/home/rkolmos/projects/templ-error-location-repro/main.templ error=/home/rkolmos/projects/templ-error-location-repro/main.templ source formatting error: 14:22: missing ',' in parameter list ]
(✗) Event handler failed [ error=failed to generate code for "/home/rkolmos/projects/templ-error-location-repro/main.templ": /home/rkolmos/projects/templ-error-location-repro/main.templ source formatting error: 14:22: missing ',' in parameter list ]
(✗) Error [ error=failed to generate code for "/home/rkolmos/projects/templ-error-location-repro/main.templ": /home/rkolmos/projects/templ-error-location-repro/main.templ source formatting error: 14:22: missing ',' in parameter list ]
(✗) Command failed: generation completed with 1 errors
```

To me it looks like what is happening is that the reported location is the location in the generated go file. It would be helpful if templ mapped that error back to it's location in the templ file similar to what the LSP does. The error is especially confusing if it is the first time the go file has been generated because there is nothing to reference to see what the offending line of code is.

Would the project be open to either?
1. Mapping errors back to their location in the templ file.
2. Providing a way to make `templ generate` keep invalid generated go files

If either of these is valuable I'd be happy to submit a PR.

joerdav:
I think that mapping the errors would be most useful. We do store map ranges of go code to templ code so it should be possible.

Robert-Kolmos:
I'm hoping to have some time to do this this week. If that changes I'll update here.

PR #737, opened 2024-05-15
fix: map errors from formatting generated files back to their locations in the templ file

Robert-Kolmos:
This is not a complete fix for the original issue #726. The generator does not run the resulting generated files through the go compiler so it is still possible to write a templ where the resulting go file doesn't compile but no error is output when running `templ generate`. I'd be happy to look into that as well but I wanted to put out a more minimal PR and get feedback first.

```
templ test() {
    for _ {
        <div>
    }
}
```

This PR makes 2 main changes
1. It attempts to map the locations in any errors that the formatter outputs back to their locations in .templ files. If this fails it falls back to reporting the unmapped error.
2. It uses a different error message to distinguish between when the reported error is in the original .templ file or the generated file.


cmd/templ/generatecmd/eventhandler.go
```diff
@@ -257,6 +263,47 @@ func (h *FSEventHandler) generate(ctx context.Context, fileName string) (goUpdat
...
+
+	// The sourceMap positions are off by one because of the package.
+	pos, ok := sourceMap.SourcePositionFromTarget(line-1, col)
+	if !ok {
+		return nil, false
+	}
+	return fmt.Errorf("%d:%d%s", pos.Line+1, pos.Col, suffix), true
+}
+
+func parseFormatterError(str string) (errLine, errCol uint32, suffix string, err error) {
```
a-h, line 280: I think it's better to inspect the error type.

https://go.dev/play/p/YR6TgOR9XeR

cmd/templ/generatecmd/eventhandler.go
```diff
@@ -221,7 +222,12 @@ func (h *FSEventHandler) generate(ctx context.Context, fileName string) (goUpdat
 
 	formattedGoCode, err := format.Source(b.Bytes())
 	if err != nil {
-		return false, false, nil, fmt.Errorf("%s source formatting error: %w", fileName, err)
+		mapped, ok := mapFormatterError(err, sourceMap)
```
a-h, line 225: As per the other comment, I think it would be better to look at the type of the error, and maybe even to rewrite the error that comes back.

A small test around that behaviour would be useful too, so we don't accidentally lose the behaviour in future work.

a-h (changes requested):
Thanks for looking at this!

cmd/templ/generatecmd/eventhandler.go
Robert-Kolmos: That is way better, thank you. I updated the code.

cmd/templ/generatecmd/eventhandler.go
Robert-Kolmos: Agreed, I'm stepping away for a bit and wanted to commit what I've got so far but I will add tests soon.

cmd/templ/generatecmd/eventhandler_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestEventHandler(t *testing.T) {
+	tests := []struct {
+		name           string
+		fileName       string
+		errorPositions []token.Position
+	}{
+		{
+			name:     "single error outputs location in srcFile",
+			fileName: "single_error.templ.error",
```
Robert-Kolmos, line 22: The `.error` suffix was necessary to prevent the test files from breaking the build because they would get pulled into any `templ generate` running from the root.

a-h (approved):
Looking good! Thanks for this. 😁

PR #745, opened 2024-05-19
feat: add JSONString and JSONScript functions, update docs, mark templ script as legacy

### LINKED ISSUE #739: proposal: add templ.JSONScript and templ.JSON functions
The page at https://templ.guide/syntax-and-usage/script-templates currently mentions the use of the `templ script` script template as the main way to pass Go data to scripts, and demonstrates the use of `onclick` attributes, which are listed as not best practice in MDN.

In the draft PR at https://github.com/a-h/templ/pull/738/files I demonstrate how to use data attributes and script elements to pass data, and following the best practice described in MDN docs.

However, I needed to create:

- A function to create a JSON string from a given set of data, so that it can be used to populate an attribute (the proposed `templ.JSON` function)
- A function to create a `<script>` element containing JSON data (the proposed `templ.JSONScript` function).

I propose:

- Updating the docs based on PR #738 so that it:
  - Demonstrates how to use attributes and script tags to pass data from backend to frontend
  - Demonstrates use of the NPM ecosystem
  - Demonstrates the use of esbuild to bundle
  - Recommends against the use of `templ script`, on the basis that the developer experience of writing JS/TS inside a TS file is better, and can more easily include intellisense etc.
- Adding the templ.JSONScript and templ.JSON functions.

Thoughts?

a-h:
Fixes #739


joerdav (approved):

jsonscript.go
```diff
@@ -0,0 +1,74 @@
+package templ
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+)
+
+type CSPContextKeyType int
```
a-h, line 10: Remove this in favour of the `templ.GetNonce` function in https://github.com/a-h/templ/pull/752

PR #759, opened 2024-05-26
fix: write logs to stderr instead of stdout, fixes #753

### LINKED ISSUE #753: Errors send to -stdout
### Observed behaviour

On template instantiation failure, `templ` writes an error report to `stdout`. When the flag `-stdout` is used, this causes the error messages to be written to the file, instead of being reported back to the user via `stderr`. 

This degrades the user experience because the feedback from the failure is hidden from the user. When `templ` is used from the command line, the only direct feedback is the exit code, which shells tend to hide from the user. When `templ` is used within most build tools, the tool will highlight the failure but give no explanation, which is generally expected to be the captured stderr output.

### Expected behaviour

When the flag `-stdout` is used, the user receives visible feedback about the error, in both the command line and most build tools.

### Suggested change

Use `stdout` for data output (instantiated templates), `stderr` for out-of-band messages such as errors and informational messages. This goes against [the position](https://github.com/a-h/templ/issues/439#issuecomment-1918875561) stated in issue #439, but it would align `templ` closer to standard POSIX conventions, improve its user experience, and simplify its integration with build tools.

a-h:


a-h:
This fixes https://github.com/a-h/templ/pull/753

cmd/templ/processor/processor.go
```diff
@@ -42,7 +42,7 @@ func shouldSkipDir(dir string) bool {
 }
 
 func FindTemplates(srcPath string, output chan<- string) (err error) {
-	return filepath.Walk(srcPath, func(currentPath string, info fs.FileInfo, err error) error {
+	return filepath.WalkDir(srcPath, func(currentPath string, info fs.DirEntry, err error) error {
```
a-h, line 45: Shouldn't stricly be in this PR, but I noticed that Walk was being used, and WalkDir is more efficient.

PR #760, opened 2024-05-26
Define tip for template generate watcher

manimovassagh:
define  small helper definition for making it easier for developers with using --watch flag


a-h (changes requested):
There are spelling (helpfull), formatting (missing line break) and grammar problems (specially) with this change.

It looks like you think it's useful to point out the `templ generate --watch` option in this section.

Would a link to that section be better? Or to use the `:::tip` feature to highlight that?

a-h (approved):

PR #765, opened 2024-05-30
feat: preserve indentation of multiline go string literals

### LINKED ISSUE #743: fmt: broken mutliline raw string formatting
templ version: v0.2.680

When formatting a multiline raw string inside of an argument the content of the string gets indented every time you format the code.

This code:
```
templ something(stuff string) {
	{ stuff }
}

templ someOtherTemplate() {
	@something(`Hi
some cool text

foo 

bar

		`)
}
```

Gets formatted to this:

```
templ something(stuff string) {
	{ stuff }
}

templ someOtherTemplate() {
	@something(`Hi
	some cool text
	
	foo 
	
	bar
	
			`)
}

```

And that code to this:

```
templ something(stuff string) {
	{ stuff }
}

templ someOtherTemplate() {
	@something(`Hi
		some cool text
		
		foo 
		
		bar
		
				`)
}

```

This repeats every time you format.

joerdav:
Fixes #743 

The solution to this was a bit odd. I thought it would be more straight forward but no!

Essentially, I allow go fmt to format the code, then I indent everything by one, then format again. If go fmt removes the line then we are safe to auto-indent, but if the new indentation is preserved then we should leave it be.

Open to other solutions.

### DIFF
--- a/parser/v2/types.go
+++ b/parser/v2/types.go
@@ -194,6 +194,7 @@ func (exp TemplateFileGoExpression) Write(w io.Writer, indent int) error {
 	return err
 }
 
+//nolint:unused // This could be useful.
 func writeLinesIndented(w io.Writer, level int, s string) (err error) {
 	indent := strings.Repeat("\t", level)
 	lines := strings.Split(s, "\n")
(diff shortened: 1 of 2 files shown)


parser/v2/types.go
```diff
@@ -194,6 +194,7 @@ func (exp TemplateFileGoExpression) Write(w io.Writer, indent int) error {
 	return err
 }
 
+//nolint:unused // This could be useful.
```
a-h, line 197: What's the deal with this? Not sure why we'd need an unused variable.

joerdav: It can be removed, I figured it could be a useful function in the future, but I suppose we could grab it from git history.

PR #774, opened 2024-06-02
add Datastar documentation

delaneyj:
Currently in the process of moving Datastar's site to Templ.  Have created helpers to make Templ the defacto way to build apps with Datastar using Go. 

Also `prismjs` was a missing dependency for the docs to build locally.

Assigning this as draft until Datastar site is updated.  Thoughts in the mean time are appreciated


joerdav:
Hey thanks for taking the time for this! I'll take a proper look tomorrow, after a very quick scan of the examples I noticed `MustJSONString` which I suppose must panic? Templ attributes actually support error values returned from functions, such as `func() (string, error)`. Like I say I haven't properly read this PR so I may be missing something, but maybe `templ.JSONString` might work here to avoid a panic? https://templ.guide/syntax-and-usage/script-templates/#pass-server-side-data-to-the-client-in-a-html-attribute

delaneyj:
@joerdav thank you so much! I didn't know about `templ.JSONString()` I've moved over the site to using that directly and this example.  The other was a stopgap, this is much better.

delaneyj:
https://data-star.dev is now at v0.15.0 and all [templating is using templ](https://github.com/delaneyj/datastar/tree/main/backends/go/site).  It's been great, still think interface/primitive auto format and a way to do UnsafeStyles instead of using `templ.Atttributes` would be great for situations like HOWL stacks where you aren't giving direct access to users.

Issue #775, opened 2024-06-03
Duplicate Import Statements in Generated Templates Cause go build Failure


akdotio:
### Issue Description:

When a user imports certain packages explicitly in a template file, the generated template may include duplicate import statements. This duplication causes `go build` to fail.

#### Affected Packages:

- `"github.com/a-h/templ"`
- `"context"`
- `"io"`
- `"bytes"`
- `"strings"`

### Issue Example:

```go
package main

import (
	"github.com/a-h/templ"
)

type LayoutProps struct {
	Content templ.Component
}

templ Layout(props *LayoutProps) {
	<div>
		@props.Content
	</div>
}
```

In this example, the generated Go code from the template will have two import statements for the package `"github.com/a-h/templ"`.

### Resolution:

The template generator should include import statements only if they are not already present in the template file.

joerdav:
Thanks for raising, this is definitely a confusing bit of templ, the transparent importing of packages in the generated files.

One thought I had while looking at #757 was maybe we should force the template author to import templ explicitly as you have done, the problem comes with the other packages, a user may be confused as to why suddenly they are required to import `strings` even though they aren't using it.

The less than ideal solution would be to hide all the std lib package usages in the `templ` package, so that generated templates only rely on `templ`. This would bloat the runtime package a bit, but would allow us to have only templ as a required import.

The benefit of this explicit importing of packages is that it is clearer what is happening, and less confusing when someone tries to import `io` and get an error. It would also make implementing the organize import action simpler, as the import sections would be identical between generated templates and the template itself.

A drawback would be that this is a breaking change, though there may be a way to make it backwards compatible. 

@a-h do you have any thoughts on this?

akdotio:
@joerdav I just opened a [PR](https://github.com/a-h/templ/pull/776) with a potential resolution for this issue. The changes aim to maintain the current behavior while preventing duplicate import statements in the generated code when users explicitly import the same packages in the template file.

I'm not entirely sure if this fix is related to https://github.com/a-h/templ/issues/757. If you could review the proposed changes and share your thoughts, it would be greatly appreciated.

Thanks!

a-h:
Thanks for this. I love that you not only saw the problem, but that you took the time to put together a solution. Much appreciated.

I haven't had chance to look into this PR properly (I'm prepping for speaking at BigSkyDevCon on Saturday), but thought I'd write down my first thoughts while they're fresh.

[...]

PR #776, opened 2024-06-03
Fix for Issue #775: Prevent Duplicate Import Statements in Generated Templates

akdotio:
This PR addresses issue #775, where duplicate import statements in generated templates caused `go build` to fail. 

#### Changes Made:

- Updated the template generator to check for existing import statements before adding new ones.
- Added tests to ensure that import statements are only included once, even if explicitly imported in the template file.

With these changes, the generator will no longer produce duplicate import statements, ensuring successful builds.


a-h:
Thanks a lot for for putting this together, but we realised that we needed to sort out the organisation of imports too, so we've implemented a fix for the imports more generally at https://github.com/a-h/templ/pull/793

It makes it so that there's only two imports - templ, and templruntime, so the generated code won't have magic imports that you can't see like `io` etc. Additionally, any missing imports will be added automatically.

PR #784, opened 2024-06-07
feat: add nonce to reload script

phenpessoa:
I realized that the relaod script tag was hard coded as a constant string. This was causing errors when using CSP with the newly added `WithNonce` function. 

I've made this PR so that the `modifyResponse` method parses the nonce from the request header (if present), and pass it along to `insertScriptTagIntoBody` to have it added to the script.


a-h (approved):

a-h:
Thanks a lot!

PR #802, opened 2024-06-22
feat: use buffered IO to enable streaming responses, and implement flush component

### LINKED ISSUE #207: runtime: Optimization Request: Improve TTFB by Streaming Rendered HTML
**Background**: The changes introduced in Issue #56 have an unintentional side effect related to a rendered template's  [TTFB (Time To First Byte)](https://web.dev/ttfb/).

**Problem**: With the latest update, the generated Go code first writes the rendered HTML to a buffer and only then writes to the final `Writer`. This means the entire template has to be rendered before any data is sent to the user. This causes the TTFB to become the time taken for the complete server-side page rendering rather than the time to generate the first portion of HTML. This delayed TTFB could negatively impact SEO scores.

**Example**:

```go
templ hello(name string) {
    <div>Hello, { name }</div>
}
```

For the above template, the first piece of HTML (i.e., `<div>`) could theoretically be sent immediately, allowing browsers to start rendering sooner. But with the current approach, nothing gets sent until the entire template, including `</div>`, is rendered and buffered.

**Request**: While the latest changes have been excellent from an ergonomic standpoint, it would be beneficial to either change the current behavior or provide an option to optimize the TTFB by allowing for streaming of the rendered HTML.

a-h:
Fixes #207
Resolves https://github.com/a-h/templ/discussions/458

See https://github.com/a-h/templ/discussions/781#discussioncomment-9680731 for discussion


a-h:
The `ComponentHandler` that's used for HTTP currently renders everything to an internal buffer, so that if there's an error during template rendering, the correct HTTP status code can be used.

So, there is likely a need for a streaming version of the `Handler` function, e.g. `StreamingHandler` which doesn't do that, or a streaming mode of operation for the existing version.

joerdav:
This makes sense to me, it took a minute for me to understand what the point was, but I get it now.

I think this needs thorough docs, not just around how to use it, but when and why you would use it with examples.

joerdav (approved):

PR #808, opened 2024-06-25
update documentation on air's capabilities

advdv:
Since air version 1.52.0 it also supports a proxy to support live reloading. This commit updates Templ's documentation to mention this so users can make a more informed description.


a-h:
Any reason why this is a draft PR? I think it's OK to add, or maybe not even worth mentioning when the feature was added to air, since the latest versions have it.

advdv:
Hmmm, i don't know to be honest. it is a while ago. It could be that i simply forgot to no longer flag it as read for review.

a-h:
Sorry, some of the files have moved around, I can't merge this PR. I've updated the docs though, hope you don't mind!

advdv:
of course, no problem!

PR #821, opened 2024-07-01
docs: fix typo in "Live reload with other tools"

leon-win:
This PR fixes typo on page https://templ.guide/commands-and-tools/live-reload-with-other-tools/#reload-event: `build.bin` instead of `build.cmd` in "Note" block.


a-h:
Ah, thanks for that! Nice spot!

Issue #798, opened 2024-06-19
Proposal: Change URL's signature to match [S]Printf


emehrkay:
Go from:

```go
func URL(s string) SafeURL {
```

To:

```go
func URL(format string, a ...any) SafeURL {
   s := fmt.Sprintf(format, a...)
```

This would be a backwards compatible change and would save some function nesting in templ funcs

a-h:
I like the idea of having a url format / builder function.

If it was the signature of `Sprintf`, e.g.: `urlf.Sprintf("/path/%s/fileName.js?%s=%v", "pathvar/", "qkey?", "&other=true")` there's a few dangers to consider... Because of the context of the variable substitution, there's a risk of injection attacks, so the values would need to be appropriately escaped:

* path variables need to be path escaped so that `/` etc. is encoded.
* query variables need to be query escaped so that `?`, `=` etc. are properly encoded.

There's also the question of what happens if you do something like:

```go
urlf.Sprintf("?%s", "data")
```

It could live in `github.com/a-h/templ/urlf` to avoid the name of oddity of `templ.URLf` or similar, but I don't mind `templ.URL`

I wonder if a builder style is better, since it turns issues with `sprintf` replacement into compile time issues:

```go
urlbuilder.Build().Path("a").Path("b").Path(c).Query(key, value)
```

joerdav:
There could be an unsafe counterpart too, `urlf.UnsafeSprintf`.

a-h:
I think the next action is to track usage of the package. If it's used a lot, then we merge it into the mainline.

This is where Github needs a "remind me in 6 months" function. 😁 

Good ideas always come back, so I'll close for now to keep the issues list as manageable as we can.

PR #836, opened 2024-07-06
feat: url builder

garrettladley:
Implementation of a URL Builder per the discussion in #798

Had 1 conceptual question, should this support mailto, tel, ftp, ftps just like `templ.URL`? If so, how do you envision the functions? Change `urlbuilder.New` to `urlbuilder.NewURL`, then add`urlbuilder.NewMailto`, etc?

### DIFF
--- a/docs/docs/03-syntax-and-usage/14-context.md
+++ b/docs/docs/03-syntax-and-usage/14-context.md
@@ -134,6 +134,10 @@ templ themeName() {
 }
 ```
 
+:::note
+As of v0.2.731, Go's built in `context` package is no longer implicitly imported into .templ files.
+:::
+
 ## Using `context` with HTTP middleware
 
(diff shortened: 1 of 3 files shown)


a-h:
This looks great. My only reluctance is to merge it into the templ mainline straight away.

Was thinking it might be a good idea to create a new module `github.com/templ-go/x` and add this package into that where there's no expectation of stable APIs, then get feedback from users, before merging into templ.

garrettladley:
I like that idea, I could go ahead and do that, just need the repo to be created in the templ-go organization

a-h:
See https://github.com/templ-go/x for experimental repo.

If you're able, I would love a PR that updates the docs to:

- Point at the new experimental repo,
- Documents the experimental URL builder in the templ docs
- Links to a Github discussion about the URL builder to collect community feedback

PR #840, opened 2024-07-08
feat: add `templ diagnose` command

a-h:
We're getting a number of reports where people are running a different version of templ than they think they are, or their environment is not set up correctly.

This PR adds a new `templ diagnose` command to assist with diagnosis of faults.

There's also a JSON mode, which is intended for use by the VS Code and Neovim extensions. The extensions should be upgraded to run this new command (`templ diagnose -json`) from within the context of the extension, thereby showing users the state of the environment that's visible to the editor, rather than the shell that they run commands at.


joerdav (approved):
Nice! I may have called it `templ check` as this is similar to other tools I've used to check your env. But I think it is fine as is.

a-h:
I think `check` sounds better, since `diagnose` is a more complex word, plus, it clashes with LSP diagnostic warnings.

I'll update it.

a-h:
Actually, maybe `info` is a better name. Check also sounds like it could be something that works on files.

joerdav:
Info is fine too, for reference the neovim equivalent command is `:checkhealth`

a-h:
OK, I think that's it.

PR #841, opened 2024-07-09
fix: resolve templruntime imported and not used

### LINKED ISSUE #834: "github.com/a-h/templ/runtime" imported as templruntime and not used
I have a small templ file that only includes scripts, after upgrading to templ v0.2.747 today I'm getting this error:
```
"github.com/a-h/templ/runtime" imported as templruntime and not used
```

You can reproduce it by running `templ generate` on this:
```go
package components

script SetObjectName(name string) {
    window.object_name = name
}
```

garrettladley:
Fixes #834

I took the fake import approach, as suggested by @a-h ([see](https://github.com/a-h/templ/issues/834#issuecomment-2213526815)).

I wrote it at the end of the file to avoid falling between the call to `g.writeImports()` [here](https://github.com/garrettladley/templ/blob/resolve-templruntime-imported-and-not-used/generator/generator.go#L122) and any imports in template nodes from the call to `g.writeTemplateNodes()` [here](https://github.com/garrettladley/templ/blob/resolve-templruntime-imported-and-not-used/generator/generator.go#L125). I did this to resolve errors of the form `failed to generate code for "path/to/file" ... imports must appear before other declarations`


generator/generator.go
garrettladley, line 1: main code change

a-h (approved):
Thanks a lot for fixing this!

PR #845, opened 2024-07-14
Add new ImportMapScriptElement

sylr:
> [!IMPORTANT]  
> This is a first draft, looking for inputs/insights.

`ImportMapScriptElement` renders a json [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) which allows to tell browsers how to resolve javascript dependencies.


importmapscript.go
garrettladley, line 1: is this not just a special case of a [`JSONScript`](https://github.com/a-h/templ/blob/main/jsonscript.go) where `type="importmap"` and there is a data is of type `ImportMap`? if so, we should simply reuse the logic here.

garrettladley (changes requested):
good stuff @sylr

importmapscript.go
sylr: The purpose of `ImportMapScriptElement` is to avoid Marshaling JSON data every render.

importmapscript.go
```diff
@@ -0,0 +1,61 @@
...
+// ImportMap represents an import map JSON object.
+type ImportMap struct {
+	Imports   map[string]string            `json:"imports"`
+	Scopes    map[string]map[string]string `json:"scopes,omitempty"`
+	Integrity map[string]string            `json:"integrity,omitempty"`
+}
+
+// NewImportMapScriptElement creates a new ImportMapScriptElement.
+// The data must be a valid JSON string.
+func NewImportMapScriptElement(id string, data string) ImportMapScriptElement {
```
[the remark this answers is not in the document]
sylr: I changed the function name as suggested.

joerdav:
Your time and effort is much appreciated @sylr ! Thanks for putting time aside to contribute a feature.

Though, we have in our guidelines an ask to first open a discussion when implementing a feature that hasn't already been discussed: https://github.com/a-h/templ/blob/main/CONTRIBUTING.md#come-up-with-a-design-and-share-it

Definitely not saying that we won't merge your change, but it would bo good to just have a proposal issue in place so that we can ensure that it aligns with the goals of templ.

Thanks again for contributing!!

PR #848, opened 2024-07-15
add primitive document symbol support

### LINKED ISSUE #347: lsp: support textDocument/documentSymbol
Hi, I am missing a feature of the symbol list(the 'Ctrl+Shift+O' menu) in your vscode extension. The result could look like this:
image
It would only have to contain the top-level `templ`, `css`, `script` and `type` declarations.

jackielii:
try to fix #347 

Currently it deals with Go symbols + parser.HTMLTemplate

TODOs:

1. deal with documentSymbol as well
2. add other symbols (css, script)?
3. add correct containerName?


jackielii:
OK, I admit there is a bit problem with my approach: I didn't convert the templ URI to the Go URI, yet somehow it worked.

It seems gopls is good at parsing partial results and skip over invalid block, therefore all the non templ elements are all returned. I then added the `HTMLTemplate`, making the result correct....

joerdav:
Is this ready to be reviewed @jackielii ?

joerdav:
I spotted the TODO, oops! Thanks for the PR by the way, it's looking to be almost there!

jackielii:
I've been using it, works most of the time. But because it uses gopls to parse the templ files directly. It has quite a few broken cases. I'll redo it using the proper way by converting to go files and find mapping symbols.

a-h:
Legend. 😀

jackielii:
@joerdav @a-h the PR is ready as is. Please feel free to review and comment.

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -825,9 +826,48 @@ func (p *Server) DocumentLinkResolve(ctx context.Context, params *lsp.DocumentLi
...
+			if err != nil {
+				return nil, err
+			}
+		}
+		switch s := s.(type) {
+		case lsp.DocumentSymbol:
+			convertRange(&s)
+			result = append(result, s)
+		case lsp.SymbolInformation:
+			// p.Log.Info("symbole range before", zap.Any("range", s.Location.Range), zap.String("uri", string(s.Location.URI)))
```
joerdav, line 862: I think these logs can be removed.

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -1216,3 +1256,24 @@ func (p *Server) Request(ctx context.Context, method string, params interface{})
 	defer p.Log.Info("client -> server: Request end")
 	return p.Target.Request(ctx, method, params)
 }
+
+func mapToSymbol(m map[string]interface{}) (interface{}, error) {
```
joerdav, line 1260: Unfortunate that we have to do this! But I can't think of a way around it either.

joerdav (changes requested):

cmd/templ/lspcmd/proxy/server.go
jackielii: Indeed, this is just how they define the protocol, the best way is to put it in a custom marshaler, however, here we have already received the data in the `map[string]any`

joerdav (approved):
Thanks for the contribution!! I've tested this out and it works exactly as I would expect it.

a-h:
Thanks a lot for seeing this through to completion. Much appreciated.

Issue #850, opened 2024-07-16
Proposal: Introduce JsGenericVar for Event and Element Access in element event handlers


cornejong:
While using temple i came acros the following issue:

When defining a `script` that will be added to the `onclick` attribute of an element, there is currently no way to access the event or the target element in the handler. This would have me end up doing additional querySelectors to access the target element and, maybe most important, all event metadata is completely lost in this way.

## Solution
Introduce a JsGenericVar string type. This type can be used to set the signature of the function and be replaced by the string value as a JS variable during generation.
``` go 
// JsGenericVar refers to a generic JS variable name.
// The string value of JsGenericVar will be used as the variable argument in the function call
type JsGenericVar string
```
This would enable access to any global JS variable in a Templ script block.

``` go
// JsGenericVar defines a generic JS variable to be replace with the string value within the Javascript function call on generation
type JsGenericVar string

script OnTheClick(event templ.JsGenericVar, element templ.JsGenericVar) {
    console.log(event, element)
}
templ TheButton() {
    <button onclick={ OnTheClick( templ.JsGenericVar("event"), templ.JsGenericVar("this") ) }>TheButton</button>
}
```
Since there is no type-checking in the script blocks (because it's JavaScript), this string type should work seamlessly.

### Generated HTML/JS Code
The generated HTML/JS code would look something like this:
``` html
<script type="text/javascript">
    function __templ_OnTheClick_985f(event, element) {
        console.log(event, element)
    }
</script>
<button onclick="__templ_OnTheClick_985f(event, this)">TheButton</button>
```
### Predefined Constants
Define some constants for the most common use cases: events and target element reference.
``` go 
const (
    JsEvent         JsGenericVar = "event"
    JsTargetElement JsGenericVar = "this"
)
```
This adds more explicit declarations, enhancing readability since the inferred JS variable is also stated in the script signature.
``` go
script OnTheClick(event templ.JsEvent, element templ.JsTargetElement) {
    console.log(event, element);
}

templ TheButton() {
    <button onclick={ OnTheClick(templ.JsEvent, templ.JsTargetElement) }>TheButton</button>;
}
```

Feedback and insights on this would be appreciated!

iambpn:
Same, I also had the same issue. It would be great to access the `event` and `this` JS objects on script function.

I am not familiar with the implementation of how script function in templ is transpiled into JS, but it seems like one generic type like `templ.JsGenericVar` or `templ.unsafeJSVar` would be enough to cover this issue. 

[...]

Issue #838, opened 2024-07-07
proposal: improve developer experience for basic passing of data to JS expressions 


mehdic213:
```
script removeFilter(attributeID string) {
	const attributesInput = document.getElementById('attributes-input');
	let attributes = attributesInput.value.split(',').filter(attr => attr !== attributeID);
	attributesInput.value = attributes.join(',');
}
```

`<button type="button" onClick={removeFilter(attr.ID)}>X</button>`

The above approach works, but using script templates is not recommended for new projects as per:
[https://templ.guide/syntax-and-usage/script-templates/#script-templates](https://templ.guide/syntax-and-usage/script-templates/#script-templates)

-------------------------------------------

```
<script>
	function removeFilter(attributeID) {
		const attributesInput = document.getElementById('attributes-input');
		let attributes = attributesInput.value.split(',').filter(attr => attr !== attributeID);
		attributesInput.value = attributes.join(',');
	}
</script>
```

`<button type="button" onclick={ fmt.Sprintf("removeFilter('%s')", attr.ID) }>X</button>`

I tried the recommanded approach (above), but when I try to use fmt to format the parameters of the function I get this error:
```
internal\template\template_templ.go:324:78: cannot use fmt.Sprintf("removeFilter(%s)", attr.ID) (value of type string) as templ.ComponentScript value in argument to templ.RenderScriptItems
internal\template\template_templ.go:332:54: cannot use fmt.Sprintf("removeFilter(%s)", attr.ID) (value of type string) as templ.ComponentScript value in variable declaration
```

- templ version v0.2.747
- go version go1.22.1 windows/amd64
- golang.org/x/tools/gopls v0.16.1

angel-git:
i have the same issue and i workarounded adding a new `data-` attribute in my button:

```js
<script>
	function removeFilter(e) { // pass element instead
                const attributeID = e.getAttribute('data-attribute-id'); // gets the new attribute in the button element
		const attributesInput = document.getElementById('attributes-input');
		let attributes = attributesInput.value.split(',').filter(attr => attr !== attributeID);
		attributesInput.value = attributes.join(',');
	}
</script>
```

```
<button type="button" data-attribute-id={ attr.ID } onclick="removeFilter(this)">X</button>
```

Nandini1071:
is this issue still open ?

Nandini1071:
if yes would like to work on this.

a-h:
[...]

PR #851, opened 2024-07-16
implemented JsExpression type

cornejong:
## Description:
This PR implements the JsExpression type for using JavaScript expressions in script template arguments, as proposed in #850 with some modifications based on discussions and insights.

## Objective:
The main objective, as described in #850, was to allow access to event objects in ```on*``` type event handlers applied on elements. This implementation enables passing any JavaScript expression to a script template.

## Usage
``` go
script OnTheClick(event templ.JsExpression, element templ.JsExpression, randomExpressionResult templ.JsExpression) {
    console.log(event, element, randomExpressionResult)
}
templ TheButton() {
    <button onclick={ OnTheClick( templ.JsExpression("event"), templ.JsExpression("this"), templ.JsExpression("1 + 2") ) }>TheButton</button>
}
```


cornejong:
Maybe, looking at it now. JsVar should be renamed to JsExpression. Since it basically would allow for any JavaScript expression to be passed this way. 

That might be a better way of describing its behavior. 

Thoughts and insights on this are very welcome.

garrettladley:
Agreed @cornejong, `JsExpression` seems to be more reflective of the functionality. Curious to hear @a-h's thoughts

joerdav:
Just for information in this PR, script templates is currently deemed as a legacy feature, not recommended as an approach anymore. https://templ.guide/syntax-and-usage/script-templates#script-templates

The reason was exactly that it wasn't flexible feasible to support all the different ways that javascript is used in modern web development. 

The recommendation now is to use script tags, or js assets to provide scripting abilities. That way we can focus on the dx of editing js within html.

I hope this helps. But it may mean that this PR will not be accepted, @a-h would you agree?

cornejong:
Thanks for the info @joerdav, Somehow i missed that (quite big) alert in the docs 😅
Although i realy like the script templates for smal simple projects, if it's considered legacy i could see why it won't be accepted. I personally would still like to see it added, but we'll see how its goes.

For now i'll just leave the PR here and let @a-h decide it's fate.

iambpn:
@joerdav yes it's true, that `script` method in templ is not recommended, but usage of `script` method is still the recommended way to handle `on*` HTML attribute like `onClick`, `onChange`  and while working with event attribute access to `this` and `event` is a must.

As you mentioned, there is a workaround with `<script>` tag and `templ.NewOnceHandle` but with its own limitation of not being able to pass data from `go` to `js` directly.

for example this would not be possible if we use `<script>` tag:
```templ
templ ShowText(txt string) {
	<div>
		<input
			type="checkbox"
			onChange={ descriptionClicked(txt) }
		/>
	</div>
}

script descriptionClicked(text string) {
	console.log(text)
}
```

joerdav:
I would disagree @iambpn I think that this is possible, but is a slightly different syntax.

If you are happy to take on a new dep you could use https://github.com/gnat/surreal

```templ
templ ShowText(txt string) {
	<div>
		<input
			type="checkbox"
			data-text={ txt }
		/>
		<script>
			me("-").on("change", evt => console.log(me(evt).attribute('data-text')))
		</script>
	</div>
}
```

This can also be done with no deps

```templ
templ ShowText(txt string) {
	<div>
		<input
			type="checkbox"
			data-text={ txt }
		/>
		<script>
			const comp = document.currentScript.closest('div')
			const checkbox = comp.querySelector('input')
			checkbox.addEventListener('change', function() {
				console.log(checkbox.dataset.text)
			})
		</script>
	</div>
}
```

iambpn:
@joerdav this is also a great workaround, but it seems little trivial to go through dataset property. I would love to have this and event access in script method, but if this is the recommended way moving forward, I have no issues.

runtime.go
```diff
@@ -556,13 +556,20 @@ func RenderAttributes(ctx context.Context, w io.Writer, attributes Attributes) (
 func safeEncodeScriptParams(escapeHTML bool, params []any) []string {
 	encodedParams := make([]string, len(params))
 	for i := 0; i < len(encodedParams); i++ {
+		// If its a JsExpression, Just leave it and don't encode it
```
a-h, line 559:
```suggestion
```

js_expression.go
```diff
@@ -0,0 +1,5 @@
+package templ
+
+// JsExpression represents a JavaScript expression intended for use as an argument for script templates.
+// The string value of JsExpression will be inserted directly as JavaScript code in function call arguments.
+type JsExpression string
```
a-h, line 5:
```suggestion
// JSExpression represents a JavaScript expression intended for use as an argument for script templates.
// The string value of JSExpression will be inserted directly as JavaScript code in function call arguments.
type JSExpression string
```

js_expression_test.go
```diff
@@ -0,0 +1,14 @@
...
+)
+
+func TestJsExpression(t *testing.T) {
+	expected := "myJsFunction(\"StringValue\",123,event,1 + 2)"
+	result := SafeScriptInline("myJsFunction", "StringValue", 123, JsExpression("event"), JsExpression("1 + 2"))
+
+	if result != expected {
+		t.Fatalf("TestJsExpression: Expected '%s' but got '%s'", expected, result)
+	}
+}
```
a-h, line 14:
```suggestion
func TestJSExpression(t *testing.T) {
	expected := "myJSFunction(\"StringValue\",123,event,1 + 2)"
	actual := SafeScriptInline("myJSFunction", "StringValue", 123, JSExpression("event"), JSExpression("1 + 2"))

	if actual != expected {
		t.Fatalf("TestJSExpression: expected %q, got %q", expected, actual)
	}
}
```

a-h (commented):
Looks great.

I've put some suggestions in to change the name from `Js` to `JS` which I think fits the Go naming conventions more.

Re: the filename, I think it should be `jsexpression.go`  to match the rest of the project.

I'd also like an example in the generator test suite, and some documentation to describe common use cases, e.g. adding an event object to handlers.

a-h:
Looking good. I think this is a good solution to some of the problems. It doesn't deal with async functions or other issues, but it does unlock the use of passing event objects etc. to the functions, and it's not a big maintenance burden.

cornejong:
@a-h Thanks!

I've committed your suggestions and chanaged the file names.
In a bit i'll add it to the generator test suite and the examples to the docs.

cornejong:
@a-h, tests and docs are added.

Out ot quriosity. You mentioned "It doesn't deal with async functions or other issues". What are currently some JS issues? I had a look in the github issues i couldn't find any open issues.

a-h:
Re: script template issues, at the moment, for example, with script templates there's no way to:

- Put multiple functions in a script template
- Write async functions
- Use TypeScript or other build processes
- Include NPM packages
- Format JS code
- Get LSP functionality for JS

a-h:
Thanks!

cornejong:
> Re: script template issues, at the moment, for example, with script templates there's no way to:
> 
> * Put multiple functions in a script template
> * Write async functions
> * Use TypeScript or other build processes
> * Include NPM packages
> * Format JS code
> * Get LSP functionality for JS

Alright, yeah I've ran into some of these a few times. Have some idea's about some of them. l'll see if i can come up with some elegant solutions in my spare time.

a-h:
If you want to catch up on a video call etc. to bounce ideas, can do that...

cornejong:
> If you want to catch up on a video call etc. to bounce ideas, can do that...

That might be nice, but I'll first tinker around a bit. See if any of these ideas hold water. 

I've already played around with a really simple/bare bones typescript addition using the "typescript" keyword in place of "script". But that leaves much to be desired.

a-h:
Take a look at https://github.com/a-h/templ/issues/838 - I've just added some ideas at the end of that, riffing on what you've already added.

PR #871, opened 2024-08-03
fix: advance past comments in raw go expressions

### LINKED ISSUE #870: Raw go errors if block ends with a comment
**Describe the bug**
Comments in the middle of a raw code block is fine, but a comment at the end will cause an error when generating templates

**To Reproduce**
```templ
// example.templ

package bingo

templ innerDataPatient() {
    {{
        one := "one"
        two := "two"
        // fine
        four := "four"
        // breaks
    }}
    <div></div>
}
```
Then run:
```bash
$ TEMPL_EXPERIMENT=rawgo templ generate
(✗) Error generating code [ file=input.templ error=input.templ parsing error: go code: missing close braces: line 8, col 8 ]
(✗) Event handler failed [ error=failed to generate code for "input.templ": input.templ parsing error: go code: missing close braces: line 8, col 8 ]
(✗) Error [ error=failed to generate code for "input.templ": input.templ parsing error: go code: missing close braces: line 8, col 8 ]
```
**Expected behavior**
A comment at the end of a raw go code block should not cause an error

 - OS: MacOS
 - templ CLI version: v0.2.7.47
- Go version: 1.22.5 darwin/arm64

**Additional context**
I suspect it's a simple parser error

knudtty:
Fixes #870 

When looking into the internal go/scanner package, I found the final comments are skipped. It works by scanning til the end of a comment, then a goto to the beginning of the Scan function if the mode is not set to ScanComments. If the scanner encounters the final closing brace, the position returned has not advanced. 

I implemented a solution that scans comments and advances the length of the token literal. While writing this description I realized the `end` variable in the go expression extractor probably just needs to be updated properly. Essentially I have implemented 2 fixes. Please review both (small) commits, in case one is a better solution than the other.


a-h:
Did you run the tests? I'm getting an error with this PR. There's also no test case to check the new behaviour.

a-h:
I've updated the PR to the test case you provided in the issue, and the first commit in this PR.

Thanks!

PR #874, opened 2024-08-04
feat: add support for lazy generation

Aerek-Yasa:
The `templ generate` command generates `.go` files from `.templ` files. When the `.go` file is newer than the `.templ` file it indicates that the `.templ` has not been updated since the `.go` file was generated. This means that the `.go` file is still valid and does not need to be generated again. *This ignores users manually editing the generated `.go` files, but there is usually not a good reason to do this.* 

This pull request contains a proposal to only generate the `.go` file if it is older than its source `.templ` file. This functionality was put behind a `--lazy` CLI flag to maintain backwards compatibility. 

~Aerek


a-h:
This looks great. Sorry for the late attention, been on holiday. 😁

a-h (approved):

PR #880, opened 2024-08-11
feat: new minify-js arg for templ generate

garrettladley:
completed version of @cornejongs draft pr [here](https://github.com/a-h/templ/pull/772)

handled `templ Component() { <script ...>...</script> }` and `script Component() { ... }` cases.

i've tested this locally by comparing outputs when -minify-js=true vs -minify-js=false with `go run ./cmd/templ generate -include-version=false -f [file] -minify-js=[bool]`

do you have any ideas / starting points on how to write tests for an argument to `templ generate`? i didn't see any other examples in the project


garrettladley:
also, does anyone know how to fix the CI/build step? running into the following error:
`generator/generator.go:20:2: cannot find module providing package github.com/tdewolff/minify/v2: import lookup disabled by -mod=vendor`

cornejong:
Honestly, i kinda forgot that i started working on that  😅
Took a look at it yesterday and today, and finished my implementation.
It differs from yours in a few ways:

- Add CSS minification as  well (why not?) ```--minify-css```
- Script tag ```src``` attribute check. No need to run the minifier on tags that load in scripts remotely.
- Script tag ```type``` attribute check. Not all script tags neccesarily contain javascript / minifiable contents.

The attribute checks are build on the assumption that those attribute is defined using a constant/static attribute declaration (not using a spread or conditional declaration). Mostly because of the way those are implemented right now. there is no easy check if those attributes are present/set at generation. I'll leave it up to @a-h to decide if that's acceptable or not. Might look into making that a bit more bomb proof in the future. For now it seems to work quite well. 

Any script tag content type that is supported by the minifier ([tdewolff/minify/v2](https://github.com/tdewolff/minify/tree/v2.20.32) ) will be minified. including Json content.

Css minification is only applied to ```<style>``` tags. The ```css SomeClass() { ... }``` syntax already seemd minified.

Issue #857, opened 2024-07-22
`templ generate` takes a long time for unformatted files


MilkeeyCat:
Try to generate code for this code 

```
package test

templ TestThingy(){
    <div>
        <h3>Dummy text</h3>
        <p>
            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>

            <span>Dummy text</span>
            <br>
        </p>
    </div>
}
```

For me it takes 6.3s when it's formatted but after I run `templ fmt` on the file, it takes 4.2ms. It doesn't seem like a big difference but I have a while which takes 1min to generate xd but after I formatted it, it takes only 20ms.

**`templ info` output**
usage: templ <command> [<args>...]

templ - build HTML UIs with Go

See docs at https://templ.guide

commands:
  generate   Generates Go code from templ files
  fmt        Formats templ files
  lsp        Starts a language server for templ files
  version    Prints the version

**Desktop (please complete the following information):**
 - OS: Linux
 - templ CLI version v0.2.747
- Go version 1.22.3 linux/amd64

a-h:
I'd guess it's taking a long time to parsenthe <br> tag. The parser currently reads to the end of the content to check if it ever gets closed.

If it gets closed and there's content in it, IIRC, the formatter leaves it alone. But if it's a void tag like br, and it's got no children, templ fmt converts it to a self closing tag which is much faster to parse.

But several seconds to do either seems like a long time!

MilkeeyCat:
Yep, when i change it to `<br/>` it works MUCH faster, thanks a lot!

joerdav:
There's possibly a way to speed this up by avoiding the rewinding and re-parsing of tags after unclosed tags. 

Though it may be one of those that adds more parsing complexity than it's worth.

a-h:
[...]

PR #887, opened 2024-08-17
refactor: improve void element parsing perf, fixes #886 and #857

### LINKED ISSUE #886: performance: templ parse takes a long time and uses high CPU when unclosed void elements are used
**Describe the bug**
With the code I show here templ takes VERY long to compile, about 29 seconds. If I add more cases to the switch case it takes even longer, to the point where it is completey unusable and locked.

If I just remove the inputs in the top of the ActionStep template it goes below one second again...

This seems like a really funkey scanner issue to me, but I am sure you are much faster in finding the actual issue... 

the templ process also climbs to ridiculous CPU usage while trying to generate in this case...

**To Reproduce**
Repo: https://github.com/s00500/templ-switchcaseissue

**Expected behavior**
Fast compile, no lockup

**Logs**
```
❯ templ generate
(!) templ version check: failed to read go.mod file: open /go.mod: no such file or directory
(✓) Complete [ updates=1 duration=28.554246583s ]
````

**`templ info` output**
```
❯ templ info
(✓) os [ goos=darwin goarch=arm64 ]
(✓) go [ location=/usr/local/go/bin/go version=go version go1.22.1 darwin/arm64 ]
(✓) gopls [ location=/Users/lb/go/bin/gopls version=golang.org/x/tools/gopls v0.16.1 ]
(✓) templ [ location=/Users/lb/go/bin/templ version=v0.2.759 ]

```
macOS (latest, sonoma 14.6.1)

❯ templ version
v0.2.759

(Masterbuild from today)

templ

### PR DESCRIPTION
**Pull request by a-h**
This PR changes the behaviour of templ's void element handling.

Previously, you could do `<input>Text</input>` and templ would allow it, despite it not being valid HTML. The LSP would emit a warning, but it would render.

However, this caused a performance issue for unclosed void elements. templ was parsing the rest of the file to look for a closing element, just in case it was present, so that the elements could be nested in templ's object model.

This operation wasn't memoized, so it ran every time a void element was encountered. In files with lots of void elements, it would cause a long delay!

The formatting operation rewrites `<br>` to `<br/>` which vastly reduces the time, but it's not a great experience that it takes a long time. It seems that some users don't use the editor extensions, or are unaware of the fmt tool, so they just live with the known bad performance.

This PR changes the templ behaviour so that it doesn't have nested content for all known void HTML elements, e.g.:

```html
<input>Text</input>
```

Becomes:

```html
<input/>Text
```

This improves the performance, see benchmark results:

## Before

```
pkg: github.com/a-h/templ/parser/v2
BenchmarkParse-16           3538            387117 ns/op          194075 B/op       4684 allocs/op
BenchmarkFormat-16          3138            400745 ns/op          197759 B/op       4765 allocs/op
```

## After

```
pkg: github.com/a-h/templ/parser/v2
BenchmarkParse-16          23718             46576 ns/op           23478 B/op        544 allocs/op
BenchmarkFormat-16         19960             60805 ns/op           26874 B/op        625 allocs/op
```

## Comparison

The important part is the ns/op figure - how many nanoseconds each operation takes.

46576 / 387117=12%, so the new version is about 8 times faster in the given scenario.


joerdav (approved):
Looks good to me, this makes a lot of sense!

Issue #859, opened 2024-07-24
Templ formatting breaks imports at v0.2.747


kanstantsin-chernik:
There is a problem we noticed with formatting after upgrading to v0.2.747:
* Some lines are duplicated with and without an alias

**To Reproduce**
To reproduce you need multiple lines with hyphen in the folder name but package name without like this
`import "someproject/components/standard-metadata"`

`someproject/components/standard-metadata/some.go` file has a `standardmetadata` package name. In go it is ok to import a package with hyphens without an alias.
`templ fmt .` adds the alias, which isn't a big deal, but breaking imports is.

**Expected behavior**
Either leave imports as is without adding the aliases or add the alias but not remove or duplicate lines

**Logs**
N/A

**`templ info` output**
Run `templ info` and include the output.
```
 % templ info
usage: templ <command> [<args>...]

templ - build HTML UIs with Go

See docs at https://templ.guide

commands:
  generate   Generates Go code from templ files
  fmt        Formats templ files
  lsp        Starts a language server for templ files
  version    Prints the version

% templ version
v0.2.747
```

**Desktop (please complete the following information):**
 - OS: Linux
 - templ CLI version (`templ version`): v0.2.747
- Go version (`go version`): go1.22.5
- `gopls` version (`gopls version`): v0.14.2

**Additional context**
Reproducing seems to be deterministic. Same input generates same broken output

kanstantsin-chernik:
Seems like this [PR](https://github.com/a-h/templ/commit/df80639a0b836a8de74c7468e90334c316e8161e) is at fault

kanstantsin-chernik:
Here is an example.
Input:
```
package somepackage

import (
	"code.uber.internal/delivery/eats-web-feed/mapper/css-class"
	twu "code.uber.internal/delivery/eats-web-feed/presentation/utils/tailwind-utils"
	tu "code.uber.internal/delivery/eats-web-feed/presentation/utils/templ-utils"
	richtextview "code.uber.internal/delivery/eats-web-feed/shared/components/base-idl/richtext"
	"code.uber.internal/delivery/eats-web-feed/shared/components/base-web-components/base-tag"
	"code.uber.internal/delivery/eats-web-feed/shared/components/quick-add-stepper"
	"thriftrw/code.uber.internal/eats/presentation/eater/mobile-web-shared/models/catalog_section"
)

templ View(input Input) {
	if input.DisplayType == catalog_section.CatalogItemDisplayTypeList {
[...]

PR #890, opened 2024-08-19
fix: duplicate import added when import path contains hyphens (#859)

a-h:


joerdav (dismissed):
Cool, this looks good to me!

a-h:
I think the issue was actually in the underlying imports repo. Now, duplicate imports aren't being added by the tooling.

I've added a unit test to capture the expected behaviour:

* If the import isn't used, it's removed by the go `imports.Process` package, since the referenced name can't be resolved.

```templ
package main

import (
	"context"
	"os"
        // Next line is removed.
	"templ/testproject/css-classes"
)

templ View() {
}

func main() {
	View().Render(context.Background(), os.Stdout)
}
```
 
* If the import is present, it's not removed:

```templ
package main

import (
	"context"
	"os"
        // Next line is retained.
	cssclasses "templ/testproject/css-classes"
)

templ View() {
	{ cssclasses.Header }
}

func main() {
	View().Render(context.Background(), os.Stdout)
}
```

* If the import is not named, the name is added:

* If the import isn't used, it's removed by the go `imports.Process` package, since the referenced name can't be resolved.

```templ
package main

import (
	"context"
	"os"
	cssclasses "templ/testproject/css-classes"
)

templ View() {
	{ cssclasses.Header }
}

func main() {
	View().Render(context.Background(), os.Stdout)
}
```

joerdav (approved):

PR #894, opened 2024-08-22
Updated generator to check for expired context

Aerek-Yasa:
The generator does not currently check if the context has expired. This pull request updates the generator so that it generators a check for an expired context. Many of the variables in the generator have names such as `templ_7745c5c3_W`. It was not known what naming convention was used for this, so the variable name `ctxErr` was used instead. 

Here is a minimal example with the current output and the expected output.

```Templ
// component.templ
package main

templ Component() {
    
}
```
```go
// main.go
package main

import (
	"context"
	"fmt"
	"io"
	"time"
)

func main() {
	comp := Component()
	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*200)
	defer cancel()
	time.Sleep(1 * time.Second)
	err := comp.Render(ctx, io.Discard)
	if err != nil {
		fmt.Println(err)
		return
	} else {
		fmt.Println("No error!")
	}
}
```
Expected output: `context deadline exceeded`.
Current output: `No error!`


generator/generator.go
```diff
@@ -403,6 +403,22 @@ func (g *generator) writeTemplate(nodeIdx int, t parser.HTMLTemplate) error {
...
+		{
+			indentLevel++
+			if _, err = g.w.WriteIndent(indentLevel, "return ctxErr"); err != nil {
+				return err
+			}
+			indentLevel--
+		}
+		if _, err = g.w.WriteIndent(indentLevel, "}\n"); err != nil {
+			return err
+		}
```
a-h, line 421:
```suggestion
		if _, err = g.w.WriteIndent(indentLevel, "if templ_7745c5c3_CtxErr := ctx.Err(); templ_7745c5c3_CtxErr != nil {\n"); err != nil {
			return err
		}
		{
			indentLevel++
			if _, err = g.w.WriteIndent(indentLevel, "return templ_7745c5c3_CtxErr"); err != nil {
				return err
			}
			indentLevel--
		}
		if _, err = g.w.WriteIndent(indentLevel, "}\n"); err != nil {
			return err
		}
```

joeaviva:
It could be worth an additional test, just to ensure that a cancelled context being passed to a component means it doesn't get rendered, and an error is returned. And then I think this is good to go from my perspective. Unless we think this behaviour warrants some documentation, I suppose it is possible that partial html responses could be returned if a request is cancelled.

joerdav:
It could be worth an additional test, just to ensure that a cancelled context being passed to a component means it doesn't get rendered, and an error is returned. And then I think this is good to go from my perspective. Unless we think this behaviour warrants some documentation, I suppose it is possible that partial html responses could be returned if a request is cancelled.

Aerek-Yasa:
A test was added that fails if `Render()` does not return an error when called with an expired `context.Context`.
Documentation was added that specifies partial output is possible when `Render()` fails. A partial rendering does not violate the general Go convention that other output parameters can not be trusted to be anything specific when the `error` output parameter is not nil.

~Aerek

a-h:
Thanks a lot for the change. Nice one!

PR #912, opened 2024-09-09
feat(lsp): add workspace folder initialization

### LINKED ISSUE #911: Neovim/Template goto definition returns error: index out of range
**Before you begin**
```
❯ templ version
v0.2.778
```

**Describe the bug**
Attempting to use Neovim templ LSP to go to the definition of a templ template inside of a .templ file will produce error:

```lua
Error executing vim.schedule lua callback: /usr/local/share/nvim/runtime/lua/vim/lsp/util.lua:17
98: index out of range
stack traceback:
        [C]: in function '_str_byteindex_enc'
        /usr/local/share/nvim/runtime/lua/vim/lsp/util.lua:1798: in function 'locations_to_items
'
        /usr/local/share/nvim/runtime/lua/vim/lsp/handlers.lua:436: in function 'handler'
        /usr/local/share/nvim/runtime/lua/vim/lsp/client.lua:687: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
Press ENTER or type command to continue
```

**To Reproduce**
https://github.com/catgoose/templ_issue_repro

Follow steps in readme

**Expected behavior**
Issuing `vim.lsp.buf.definition` on a templ template reference in a templ file should go to that templates definition

**Screenshots**
image

Taken from the reproduction repo above

**Logs**
Log file `templ.log` included in repro repo

**`templ info` output**
```bash
❯ templ info
(✓) go [ location=/usr/local/go/bin/go version=go version go1.23.1 linux/amd64 ]
(✓) gopls [ location=/home/jtye/go/bin/gopls version=golang.org/x/tools/gopls v0.16.2 ]
(✓) templ [ location=/home/jtye/go/bin/templ version=v0.2.778 ]
```

 - OS: Ubuntu 24.04.1
 - templ version: v0.2.778
- Go version: go version go1.23.1 linux/amd64
- `gopls` version: golang.org/x/tools/gopls v0.16.2

**Additional context**
This seems to have been fixed here https://github.com/a-h/templ/issues/893 but the issue persists
Reported here https://github.com/catgoose/templ-goto-definition/issues/1

tris203:
Fixes #911 

- Added logic to walk through workspace folders and initialize templ files.
- Logs warnings when source maps are not found in cache.
- Parses templates and sets source map cache contents.
- Handles errors during file reading, template parsing, and generation.

works on #911

### DIFF
--- a/cmd/templ/lspcmd/proxy/server.go
+++ b/cmd/templ/lspcmd/proxy/server.go
@@ -53,6 +53,8 @@ func NewServer(log *zap.Logger, target lsp.Server, cache *SourceMapCache, diagno
 	}
 }
 
+var preLoadURIs = []*lsp.DidOpenTextDocumentParams{}
+
 // updatePosition maps positions and filenames from source templ files into the target *.go files.
 func (p *Server) updatePosition(templURI lsp.DocumentURI, current lsp.Position) (ok bool, goURI lsp.DocumentURI, updated lsp.Position) {
 	log := p.Log.With(zap.String("uri", string(templURI)))
@@ -240,7 +242,7 @@ func (p *Server) Initialize(ctx context.Context, params *lsp.InitializeParams) (
 				return err
 			}
 			uri := uri.URI("file://" + path)
-			isTemplFile, _ := convertTemplToGoURI(uri)
+			isTemplFile, goURI := convertTemplToGoURI(uri)
 			if isTemplFile {
 				b, err := os.ReadFile(path)
 				if err != nil {
@@ -270,6 +272,16 @@ func (p *Server) Initialize(ctx context.Context, params *lsp.InitializeParams) (
 				// Set the Go contents.
 				p.GoSource[string(uri)] = w.String()
 
+				didOpenParams := &lsp.DidOpenTextDocumentParams{
+					TextDocument: lsp.TextDocumentItem{
+						URI:        goURI,
+						Text:       w.String(),
+						Version:    1,
+						LanguageID: "go",
+					},
+				}
+
+				preLoadURIs = append(preLoadURIs, didOpenParams)
 			}
 			return nil
 		})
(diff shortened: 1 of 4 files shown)


tris203:
This PR preloads the templ files into sourcemaps

This works, but the issue still persists, will need to look closer as to why

tris203:
Okay, so before the file is open, when we forward the request to gopls, it returns a result that is off by 1 (in the line only, character is fine)

This throws our mapping off. I have no idea why that would be the case at the moment

tris203:
Okay, this works.

We store the loaded URIs and their contents and then fire them into didOpen once the servers are all initialized.

I don't know if I like this implementation, but it works

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -50,6 +53,8 @@ func NewServer(log *zap.Logger, target lsp.Server, cache *SourceMapCache, diagno
 	}
 }
 
+var preLoadURIs = []*lsp.DidOpenTextDocumentParams{}
```
joerdav, line 56: Could this be a field on the `Server` struct?

tris203: Yeah. I can refactor that no problem.

cmd/templ/lspcmd/proxy/server.go
```diff
@@ -228,6 +234,61 @@ func (p *Server) Initialize(ctx context.Context, params *lsp.InitializeParams) (
...
 
+	for _, c := range params.WorkspaceFolders {
+		path := strings.TrimPrefix(c.URI, "file://")
+		werr := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			uri := uri.URI("file://" + path)
+			isTemplFile, goURI := convertTemplToGoURI(uri)
+			if isTemplFile {
```
joerdav, line 245: Sorry for not spotting this last time, it's a small nit, but I think inverting this if will improve readability and reduce indentation.

joerdav:
Happy to add this changes if you don't have time, let me know! And thankyou for the contribution and deep dive into this bug!

cmd/templ/lspcmd/proxy/server.go
tris203: Agreed, done.

tris203:
> I'm going to take some time to test this out today in various projects to ensure it works as expected.
> 
> I also wonder if we can lock this bug away with an lsp test: [`60631f9`/cmd/templ/lspcmd/lsp_test.go](https://github.com/a-h/templ/blob/60631f964f1c347f3ef5ff34caab3e01b968855c/cmd/templ/lspcmd/lsp_test.go)

I agree this should have a test, I am just not sure of the best way to approach it.

I think it would be a new templ file in the 

> I'm going to take some time to test this out today in various projects to ensure it works as expected.
> 
> I also wonder if we can lock this bug away with an lsp test: [`60631f9`/cmd/templ/lspcmd/lsp_test.go](https://github.com/a-h/templ/blob/60631f964f1c347f3ef5ff34caab3e01b968855c/cmd/templ/lspcmd/lsp_test.go)

I have added a test (and fixtures to provide for this) have a look and see if you think this is good enough

joerdav:
Tested this on my own projects and works as expected. Thanks!!

Issue #788, opened 2024-06-11
proposal: add `templ.Join(components ...templ.Component) templ.Component`


antran22:
Idea: We want to pass in a list of components as props and spread it inside a templ, similar to how we spread `{ children ...}`

**Reason:** Sometimes `children` is not enough when we need to define multiple children rendering point, especially on complex wrapper components. At this point, we will define something similar to a render props from React.

**Example:**
```
templ Modal(actions []templ.Component) {
   <div>
       <div class="main-content">
             {...children}
       </div>
       <div class="actions">
             <!-- This is the point where we want to spread the props -->
             {...actions}
       <div>
   <div>
}

templ ActionButton1() {
    <button>Submit</button>
}


templ ActionButton2() {
    <button>Cancel</button>
}

templ Main() {
   @Modal([]templ.Component{
          ActionButton1(),
          ActionButton2(),
   }) {
        <div>Content</div>
   }
}

```

This currently doesn't work. Current workaround is to iterate through `actions` directly.

```
templ Modal(actions []templ.Component) {
   <div>
       <div class="main-content">
             {...children}
       </div>
       <div class="actions">
             for _, action := range actions {
                   @action
             }
       <div>
   <div>
}
```

This proposal will improve the succinctness of the template.

**Suggested implementation:**

We can define a `CombinedComponents` type that implement `templ.Component` and represent a slice of `templ.Component`. Rendering will proceed by iterating over each children component. 

Then at those spread point we instead render a wrapped CombinedComponents.

a-h:
This strikes me as a `map` function in Go, so you could implement it as a function like this:

```go
func Chain(components ...templ.Component) templ.Component {
  return templ.ComponentFunc(func(ctx context.Context, w io.Writer) (err error) {
    for _, c := range components {
      if err = c.Render(ctx, w); err != nil {
        return err
      }
    }
    return nil
  })
}
```

And use it in templ like:

```templ
templ Modal(actions []templ.Component) {
   <div>
       <div class="main-content">
             {...children}
       </div>
       <div class="actions">
             @mypackage.Chain(actions...)
       <div>
   <div>
}
```

As an aside, if was designing it today, I wouldn't have added `{ ...children }` as a special case, but would have created a `@templ.Children()` function instead - to get the children from the context, and rendered them out.

I was initially enthusiastic about adding something new to templ's namespace, but I'm not sure it's a good thing to hide the fact it's doing a loop just to save two lines of code.

[...]

PR #929, opened 2024-09-23
feat: templ.Join method renders multiple components into a single component

Quarkz14:
Heyoo, 
I have a PR for the proposal suggested in this [issue](https://github.com/a-h/templ/issues/788) I followed the example solution you given @a-h. I have updated the docs too to explain the new Join method with an example in the `Temple Composition` section. Let me know what you think. I will rename the commit message as part of any changes in the feedback

### DIFF
--- a/docs/docs/03-syntax-and-usage/10-template-composition.md
+++ b/docs/docs/03-syntax-and-usage/10-template-composition.md
@@ -259,16 +259,8 @@ func main() {
   
   actions := []templ.Component{hello, world}
  
- templ showAll() {
-    @Modal() {
-       <div>Inserted from the top</div>
-    }
-  } 
- templ Modal(actions []templ.Component) {
+  templ Modal(actions []templ.Component) {
    <div>
-    <div class="main-content">
-     {...children}
-    </div>
     <div class="hello-world">
      templ.Join(actions...)
     <div>
@@ -277,11 +269,6 @@ func main() {
 }
 ```
 ```html title="output"
-<div id="wrapper">
- <div>
-  Inserted from the top
- </div>
-</div>
 <div class="hello-world">
  <div>
   Hello
--- a/docs/docs/03-syntax-and-usage/10-template-composition.md
+++ b/docs/docs/03-syntax-and-usage/10-template-composition.md
@@ -235,8 +235,11 @@ func main() {
 	<p>Dynamic contents</p>
 </div>
 ```
+
 ## Joining Components
-Components can be aggregated into a single Component using `templ.Join`
+
+Components can be aggregated into a single Component using `templ.Join`.
+
 ```templ
 package main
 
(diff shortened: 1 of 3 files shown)


join.go
```diff
@@ -0,0 +1,18 @@
+package templ
+
+import (
+	"context"
+	"io"
+)
+
+// Pass any number of templ.Components to get a single templ.Component with the components rendered
```
joerdav, line 8: Go function comments should begin with the name of the documented function.

```suggestion
// Join returns a single `templ.Component` that will render provided components in order.
// If any of the components return an error the Join component will immediately return with the error.
```

docs/docs/03-syntax-and-usage/10-template-composition.md
```diff
@@ -235,6 +235,49 @@ func main() {
 	<p>Dynamic contents</p>
 </div>
 ```
+## Joining Components
+With your templ components you can use `templ.Join` to generate a single `templ.Component` from your components.
+```templ
```
joerdav, line 240: Tabs not spaces ;)

Also I think that joining generated templ components here would be more intuitive for the docs:

templates.go
```templ
package main

templ hello() {
	<span>hello</span>
}

templ world() {
	<span>world</span>
}

templ helloWorld() {
	@templ.Join(hello(), world())
}
```

main.go
```go
package main

import (
	"context"
	"os"
)

func main() {
	helloWorld().Render(context.Background(), os.Stdout)
}
```

Output:
```html
<span>hello</span><span>world</span>
```

docs/docs/03-syntax-and-usage/10-template-composition.md
```diff
@@ -235,6 +235,49 @@ func main() {
 	<p>Dynamic contents</p>
 </div>
 ```
+## Joining Components
+With your templ components you can use `templ.Join` to generate a single `templ.Component` from your components.
```
joerdav, line 239:
```suggestion
Components can be aggregated into a single Component using `templ.Join`.
```

join_test.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+		return nil
+	})
+	world := templ.ComponentFunc(func(ctx context.Context, w io.Writer) error {
+		if _, err := io.WriteString(w, "World"); err != nil {
+			t.Fatalf("failed to write string: %v", err)
+		}
+		return nil
+	})
+	components := []templ.Component{hello, world}
```
joerdav, line 26: These definitions would probably be cleaner defined in the table test directly.

joerdav (commented):
Thanks for the contribution and time!! A few small nits here and there I spotted.

join_test.go
Quarkz14: I initialise them in the test structs now.

a-h (approved):

PR #931, opened 2024-09-24
First link in README.md

wdscxsj:
Without this fix, the generated [godoc](https://pkg.go.dev/github.com/a-h/templ) is polluted with links.


joerdav:
Great spot, thankyou

PR #932, opened 2024-09-24
feat(lsp): Support gotoDefinition from go->templ

### LINKED ISSUE #387: lsp: Support goToDefinition from a Go file to a templ file

tris203:
Support goToDefinition from a go file to a templ file.

Steps to test:
1. Attach both gopls and templ lsp to a main.go buffer 
2.  GoToDefinition
3. Profit

Caveats, because gopls and templ are attached, both references will be returned. Editor settings can be used to set preference to select from a list or go to the preferred option (probably the templ file)

When and if this was merged, it would need further adjustments to neovim/lspconfig to attach templ to go files too and would probably need a change in the vscode extension for the same

Closes #387

### DIFF
--- a/cmd/templ/lspcmd/lsp_test.go
+++ b/cmd/templ/lspcmd/lsp_test.go
@@ -329,7 +329,7 @@ func TestDefinitions(t *testing.T) {
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	log, _ := zap.NewProduction()
+	log := slog.New(slog.NewJSONHandler(os.Stderr, nil))
 
 	ctx, appDir, _, server, teardown, err := Setup(ctx, log)
 	if err != nil {
--- a/cmd/templ/lspcmd/lsp_test.go
+++ b/cmd/templ/lspcmd/lsp_test.go
@@ -323,6 +323,126 @@ func TestHover(t *testing.T) {
 	}
 }
 
+func TestDefinitions(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	log, _ := zap.NewProduction()
+
+	ctx, appDir, _, server, teardown, err := Setup(ctx, log)
+	if err != nil {
+		t.Fatalf("failed to setup test: %v", err)
+		return
+	}
+	defer teardown(t)
+	defer cancel()
+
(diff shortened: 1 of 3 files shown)


garrettladley (approved):

joerdav:
Sorry for a delay on this one. This is a great addition, many have said a pain point is gotodefinition from go to templ and this is the first step!

> Editor settings can be used to set preference to select from a list or go to the preferred option (probably the templ file)

I wonder if we should add the docs for how to do the above along side this change?

a-h:
I looked at this the other day, and couldn't get it to work. Probably because I hadn't told my editor to use the templ LSP for Go files too, so I'd say we do need to do that.

tris203:
I have done it in my config for neovim, when it comes to other editors:

Neovim: lspconfig change
VSCode: I assume an extension change

helix/emacs, will need PRs to the relevant repos

Am i missing anything?

joerdav:
I think that's correct yes. If we just update those things and there are no user config changes required apart from updating configs then maybe it's okay to merge this now at this point? Then we can gradually update the extensions etc?

tris203:
Is there any blockers for this, or anything further I can do to help move this along?

Napam:
> I have done it in my config for neovim, when it comes to other editors:
> 
> Neovim: lspconfig change VSCode: I assume an extension change
> 
> helix/emacs, will need PRs to the relevant repos
> 
> Am i missing anything?

Just wondering, so with this approach one uses templ lsp for everything go, or does some dynamic handling to detect "if templ detected then templ lsp else gopls"?

joerdav:
@Napam The templ lsp itself connects to a version of gopls, so you don't have to tell your editor to run both explicitly. The "if templ ..." stuff actually happens within templ itself. Therefore, with this PR if you configure neovim to run templ lsp for .go files it will handle it. There's nothing stopping you running both, but you might get multiple results for some actions.

a-h:
Thanks a lot for your work on this. I eventually got around to thinking about this, and put together two PRs at:

- https://github.com/a-h/templ/pull/1384
- https://github.com/templ-go/templ-vscode/pull/101

I don't think I would have got to it without your nudge.

PR #935, opened 2024-09-26
parser: fix memory leak in goexpression.Func (fixes #916)

### LINKED ISSUE #916: Taking too much memory with (relatively) larger files
**Describe the bug**
I'm not sure if this is normal/expected or a known issue so sorry in advance about my ignorance.
When I try to `templ generate` a single (programmatically generated) 6.4MiB .templ file which contains 4048 `templ` definitions each one is an svg from [Game Icons](https://github.com/delacannon/game-icons-inverted), it takes all the available memory and the laptop starts lagging and even if I leave it like this for a long time it won't finish and the memory consumption will lower sometimes to like 2GB then go back to taking all the available memory..
Also the LSP(vscode with the templ extencion) does the same thing when I open the same file in vscode.

**To Reproduce**
Clone [this repo](https://github.com/sudoCss/templ-mem-issue-example) and run `templ generate` in it.

**Expected issue cause**
I'm not familiar with the templ codebase but I think it may be in issue in a parser of some kind (may be specific to SVG parsing).

**Screenshots**
templmem

**`templ info` output**
```sh
❯ templ info
(✓) go [ location=/usr/bin/go version=go version go1.23.1 linux/amd64 ]
(✓) gopls [ location=/home/sudocss/.go/bin/gopls version=golang.org/x/tools/gopls v0.16.2 ]
(✓) templ [ location=/home/sudocss/.go/bin/templ version=v0.2.778 ]
```

**Desktop:**
 - OS: Linux(EndeavourOS)
 - templ CLI version: v0.2.778
- Go version: go1.23.1 linux/amd64
- `gopls` version: golang.org/x/tools/gopls v0.16.2

utrack:
Do not let whole contents of `src` escape to heap.
Fixes #916

### DIFF
--- a/parser/v2/goexpression/parse.go
+++ b/parser/v2/goexpression/parse.go
@@ -256,7 +256,7 @@ func Func(content string) (name, expr string, err error) {
 			err = errors.New("parser error: function identifier")
 			return false
 		}
-		expr = src[start:end]
+		expr = strings.Clone(src[start:end])
 		name = fn.Name.Name
 		return false
 	})


a-h:
Thanks for the research you put into this, which was presumably a lot more effort than the 1 line PR looks. 😁

PR #941, opened 2024-09-29
docs: Add --yes to npx commands

theothertomelliott:
When setting up a new repo using these instructions, I was initially confused because the Tailwind css wasn't being updated when started with `make live`. Hitting Ctrl+C was also hanging due to the same wait.

Adding the `--yes` flag allowed the commands to proceed if installation was required. I didn't add any additional explanation, but happy to include some elaboration if preferred.


a-h (approved):

a-h:
Thanks, you're right. The commands work OK if you already have a `package.json` and have run `npm i`, but not if you're starting from scratch, and that isn't clear in the docs.

PR #959, opened 2024-10-13
fix: example flake syntax error

Oudwins:
The example flake had at least two syntax errors which have been fixed. The example flake cannot be ran by itself so there might be more syntax errors.

I can replace it with a flake that actually works and can just be copied and pasted into a project if that is preferred.

The syntax errors were that the parenthesis were in the incorrect place.


joerdav:
Great spot! Thanks

Oudwins:
No worries took me embarrassingly long to realize....

PR #966, opened 2024-10-24
cmd/templ/lspcmd/proxy: append ';templ' in initialize client info

hyangah:
This allows gopls to recognize gopls instances serving for templ and count separately in go telemetry data processing.


a-h:
Ah, good idea! I'm not familiar with the format of adding a `;` followed by the name, is that common practice?

Do you any links to the telemetry dataset etc. or conversations about that?

The reason I ask is that once it's set, it's probably hard to change, since it would affect historic data collection, so I just want to cover that before merging.

I'd like to see telemetry of the templ LSP. For example, it may be possible to see new issues early using the telemetry.

hyangah:
Hi @a-h
https://github.com/golang/go/issues/70205 is the upstream issue. After the discussion, we think just setting "templ" instead of appending it is the right approach. Updated the PR.

a-h:
Thanks! I'll probably be able to look at this the day after tomorrow.

a-h:
@hyangah what's the status on this now then? I think we're still waiting on the discussion in https://github.com/golang/go/issues/70205 to resolve. I've added a comment there to explain how templ works etc.

PR #971, opened 2024-10-27
feat(storybook): components can have additional stories

schjan:
With this changes I was able to:

1. Create additional stories for a component with different Args (see also `_example/storybook.go`)
2. Using the `WithAdditionalPreviewJS(content string)` fn to bundle TailwindCSS during `npm run build-storybook`


a-h:
Thanks for this!

PR #972, opened 2024-10-27
Add proxyopen flag

quinn:
This adds a flag so that you can add `--proxyopen=false` to disable automatically opening the proxy url in a new browser window when the watcher starts. 

I tried to make it so the current behavior is the default, and you can opt into disabling it.

### DIFF
--- a/cmd/templ/generatecmd/cmd.go
+++ b/cmd/templ/generatecmd/cmd.go
@@ -374,8 +374,10 @@ func (cmd *Generate) StartProxy(ctx context.Context) (p *proxy.Handler, err erro
 			)
 			time.Sleep(d)
 		}
-		if err := browser.OpenURL(p.URL); err != nil {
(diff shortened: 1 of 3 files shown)


kvenn:
Yesss need this! Looks great

a-h:
Looks good to me. Happy to merge this when you're ready.

joerdav:
Sorry, can I just check what the intended difference here is to the `--open-browser` flag?

a-h:
Oh nice spot, I missed that @joerdav! 🤦🏻

quinn:
my bad @joerdav thanks for pointing that out! I missed it in the code too 😢

quinn:
added a doc PR: https://github.com/a-h/templ/pull/976

there's another doc page that mentions it, but this is the only one I could find when reading about the watch feature.

PR #974, opened 2024-10-29
fix: guard the generated context variable name

nderjung:
This PR fixes an issue where the variable name `ctx` in the generated code is not guarded by the typical prefix `templ_7745c5c3_`.  Unlike other variables which are generated by the associated generator methods, `ctx` is left out.  This can cause an implicit context to exist within each templ method and causes issues when nested methods are used as it is continiously overwritten.


nderjung:
Nevermind -- https://templ.guide/syntax-and-usage/context/

PR #976, opened 2024-10-30
Add comment about open-browser flag

quinn:
I couldn't find any reference to this flag. I see now there's another doc that has it, but that didn't come up when I was googling. Having it on this doc too will help I think.


a-h:
Thanks!

PR #986, opened 2024-11-03
Check path under windows to fix command `template` unknown

surfmuggle:
If you encountered something like command `template` unknown make sure to check your path environment variable.

### DIFF
--- a/docs/docs/02-quick-start/02-creating-a-simple-templ-component.md
+++ b/docs/docs/02-quick-start/02-creating-a-simple-templ-component.md
@@ -34,7 +34,7 @@ templ hello(name string) {
 
 ## Generate Go code from the templ file
 
-Run the `templ generate` command.
+Run the `templ generate` command. If it fails under windows run `where templ.exe` to find out the path of `templ.exe` and make sure that it is on your `PATH` environment variable (see `echo %PATH%`)
 
 ```sh
 templ generate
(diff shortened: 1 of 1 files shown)


a-h:
Thanks for thinking about how to make templ more accessible for Windows users.

Ensuring that your shell can find a binary on the `$PATH` or `%PATH%` variable isn't unique to Windows. It's just a standard part of software development.

I don't think that Windows-specific stuff should be in the main body of the text. Perhaps a tip might be useful?

```
:::tip
Windows: Use `templ.exe` in Powershell and Windows Command Line, use `templ.exe` in WSL.
:::
```

Then, add a link to a troubleshooting guide for finding a binary on the path in the FAQ:

```
# command not found: templ

If running `templ` gives the error:

command not found: templ

You will need to add templ to your path. See guide:

- Linux: <link>
- MacOS <link>

# <WINDOWS_ERROR_MESSAGE>

If running `templ.exe` gives the error...
```

surfmuggle:
I added the hint and removed the previous line: [PR 986](https://github.com/a-h/templ/pull/986/files).
This [FAQ](https://github.com/surfmuggle/templ/blob/main/docs/docs/15-faq/index.md) has so few lines that i assume there must be some additional files somewhere that are loaded / added. So i am hesitant to add anything to the linked FAQ.

Issue #977, opened 2024-10-31
Race Condition with rendering parts in goroutines with templ-initialized context


kanstantsin-chernik:
In the current implementation, there is an assumption that all the parts are rendered in a single thread. Context value for children is modified for the whole context. If snippets need to be rendered in goroutines, it creates a race condition where children go missing or sporadically added

**To Reproduce**
https://github.com/kanstantsin-chernik/templ/blob/suspense-async-rendered/examples/suspense-async-rendered/main.templ

I also added comments to the generated code

**Expected behavior**
Each templ component is independent and can be rendered in a goroutine

**Screenshots**
<img width="281" alt="Screenshot 2024-11-01 at 1 31 38 PM" src="https://github.com/user-attachments/assets/29e2fe1e-a7da-45c3-9909-426ae503b7bd">

**Logs**
```
WARNING: DATA RACE
Read at 0x00c01e4a13f0 by goroutine 1053:
  github.com/a-h/templ.GetChildren()
      external/com_github_a_h_templ/runtime.go:78 +0x186

Previous write at 0x00c01e4a13f0 by goroutine 1054:
  github.com/a-h/templ.ClearChildren()
      external/com_github_a_h_templ/runtime.go:68 +0x44
```

**`templ info` output**
 % templ info
(✓) os [ goos=linux goarch=amd64 ]
(✓) go [ location=/home/user/go-code/bin/go version=go version go1.23.2 X:nocoverageredesign linux/amd64 ]
(✓) gopls [ location=/opt/go/path/bin/gopls version=golang.org/x/tools/gopls v0.14.2
    golang.org/x/tools/gopls@v0.14.2 h1:sIw6vjZiuQ9S7s0auUUkHlWgsCkKZFWDHmrge8LYsnc= ]
(✓) templ [ location=/home/user/go-code/bin/templ version=v0.2.778 ]

**Desktop (please complete the following information):**
 - OS: all
 - templ CLI version (`templ version`):  v0.2.778
- Go version (`go version`): go version go1.23.2 X:nocoverageredesign linux/amd64
- `gopls` version (`gopls version`):golang.org/x/tools/gopls v0.14.2

**Additional context**

kanstantsin-chernik:
A couple of questions:
* Should Templ support this example by not reusing context or sync rendering is by design?
* If it is by design, can we at least introduce something like WithoutContext to reset it before we launch go routines?

a-h:
[...]

PR #989, opened 2024-11-05
feat: `ResetContext` func to allow resetting initialized context

kanstantsin-chernik:
The feature allows a context to be reset before reusing it in multiple go-routines ([see the issue](https://github.com/a-h/templ/issues/977)). 

An alternative solution would be to always create a new context, which would be overkill for 99.9% of single-threaded use cases.


garrettladley (approved):

joerdav:
My concern on this one is the unexpected effects that reseting the context could cause. 

- Children
- Scripts
- CSS
- Once
- Security Nonces

We should either mitigate these in some way with this solution, or document the possible outcomes.

Issue #968, opened 2024-10-25
refactor: remove a-h/pathvars library


lhorie:
**Describe the bug**
Currently, templ depends on a-h/path-vars, which doesn't have a license file. This means that even though templ itself is MIT licensed, it cannot be consumed by organizations that have strict policies around licensing/SBOM because introducing templ to a codebase would also introduce unlicensed transitive dependencies.

**To Reproduce**
Go to https://github.com/a-h/pathvars and notice there's no LICENSE file

**Expected behavior**
Dependencies of templ need to have clear licenses

**Screenshots**
N/A

**Logs**
N/A

**`templ info` output**
N/A

**Desktop (please complete the following information):**
 - OS: MacOS Sonoma 14.6.1
 - templ CLI version (`templ version`): v0.2.778
- Go version (`go version`): go1.23.2

**Additional context**
A tentative fix is proposed here: https://github.com/a-h/pathvars/pull/3

a-h:
Thanks for picking that up. Just an oversight.

I expect we can remove pathvars from templ now that standard library routing supports the same functionality.

a-h:
OK, I'll take a PR for this one. Let's find a-h/pathvars and remove it. The standard library router should provide all of the path variable support we need now.

a-h:
Closed in #999

PR #999, opened 2024-11-18
refactor: removed all references to github.com/a-h/pathvars

ninesl:
Completely removed the a-h/pathvars dependency to avoid potential issues from a missing license file. See #968.
It was currently only used in the storybook package, so I added parsing logic to achieve the same functionality for storybook's case.
A better solution could involve reworking the Storybook routes to use the new routing updates in 1.22.


a-h (approved):

a-h:
Thanks, always good to remove a dependency that's not needed.

PR #1002, opened 2024-11-20
docs: correct documentation for source-map-visualisations argument

matthewspivey:
The documentation of the `templ generate` command refers to the `-sourceMapVisualisations` argument. Using it generates the following error.

```bash
$ templ generate -sourceMapVisualisations .
flag provided but not defined: -sourceMapVisualisations
```

However, using the more conventionally named `-source-map-visualisations` argument successfully generates HTML files to visualize the templ code and its corresponding Go code, as shown below.

```bash
$ templ generate -source-map-visualisations
(✓) Complete [ updates=302 duration=520.239583ms ]
```

To resolve this, we can update the documentation to refer to the `-source-map-visualisations` argument instead.

### DIFF
--- a/cmd/templ/main.go
+++ b/cmd/templ/main.go
@@ -149,7 +149,7 @@ Args:
   -stdout
     Prints to stdout instead of writing generated files to the filesystem.
     Only applicable when -f is used.
-  -sourceMapVisualisations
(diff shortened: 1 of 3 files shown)


a-h (approved):
Thanks!

PR #1012, opened 2024-12-08
chore: update generated code to return nil instead of generated value

### LINKED ISSUE #991: request: update generated code to return nil instead of a nil value
**Describe the bug**
It's not really a bug, but more of a "thing to do, when coding in Go"

In the generated code, there is a lot of:
```
			return templ_7745c5c3_Err
		}
		return templ_7745c5c3_Err
```

but it is more idiomatic to return just nil, like this:
```
			return templ_7745c5c3_Err
		}
		return nil
```

guilhas07:
I believe this is enough to solve #991. Let me know if I missed something!

Closes #991

### DIFF
--- a/benchmarks/templ/template_templ.go
+++ b/benchmarks/templ/template_templ.go
@@ -110,7 +110,7 @@ func Render(p Person) templ.Component {
 		if templ_7745c5c3_Err != nil {
 			return templ_7745c5c3_Err
 		}
-		return templ_7745c5c3_Err
+		return nil
 	})
 }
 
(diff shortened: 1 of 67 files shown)


a-h:
I haven't merged this because I haven't been able to check out the branch and test it.

```
0 0 /Users/adrian/github.com/a-h/templ % gh pr checkout 1012
From github.com:a-h/templ
 ! [rejected]        refs/pull/1012/head -> guilhas07/main  (non-fast-forward)
failed to run git: exit status 1
```

Not sure what's going on with that. I'll try to do a cherry pick for the main file so you get "contributor" status.

a-h:
Merged in #1025 - I cherry picked your commit etc.

PR #1026, opened 2024-12-27
fix(breaking): trigger rebuild on changes to *.go files, fixes #646

### LINKED ISSUE #646: cmd: rebuild and refresh when a go file is changed in `--watch` mode
I have tried the example given in the docs: https://templ.guide/commands-and-tools/hot-reload

When trying to comment-out for example the "/" route, and then saving the `main.go` file, I don't see any reloads in my terminal and the route also still exists (when I refresh in the browser it still sends me the page).

I didn't find any related issues here and it's unclear to me how to debug this since no error or anything is logged.

The only thing that maybe causes this (given that it works for others (?)) is, that I am developing on NixOS.

I am using the following `flake.nix`:

```nix
{
  description = "Go proj flake";

  inputs.nixpkgs.url = "github:nixos/nixpkgs";  
  inputs.flake-utils.url = "github:numtide/flake-utils";

  outputs = { self, nixpkgs, flake-utils, ... }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        pkgs = import nixpkgs { inherit system; };
      in
      {
        devShells.default = pkgs.mkShell
          {
            buildInputs = with pkgs;[
                go
                templ
                gcc
            ];
          };
      }
    );
}
```

I use it via `nix develop` and run the given example command: `templ generate --watch --proxy="http://localhost:8080" --cmd="go run ."`

a-h:


a-h:
Closes #646

joerdav:
Nice! Will take a look at this Monday.

cmd/templ/generatecmd/main_test.go
```diff
@@ -42,3 +43,75 @@ func TestGenerate(t *testing.T) {
...
+			matches: false,
+		},
+	}
+	wpRegexp, err := regexp.Compile(defaultWatchPattern)
+	if err != nil {
+		t.Fatalf("failed to compile default watch pattern: %v", err)
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			test := test
```
joerdav, line 110: You shouldn't need this `test := test` anymore. Old habits die hard!

joerdav (changes requested):
LGTM! 

Does this warrant a mention in https://github.com/a-h/templ/blob/main/docs/docs/09-developer-tools/03-live-reload.md ?

cmd/templ/generatecmd/main_test.go
a-h: I still get a linter warning. I think it's because the `go.mod` sets the Go version at 1.21.

Screenshot 2024-12-30 at 13 23 54

However, it's OK for us to bump up to 1.22 now, as per https://endoflife.date/go

a-h: I'll do that in a separate change though, so it hits the release notes.

joerdav: Cool, all good then! I think once we bumb `go vet` will point out where we are doing it for no reason.

joerdav (approved):

PR #1027, opened 2024-12-27
feat: generate same dev and prod code, fixes #700

### LINKED ISSUE #700: refactor: update dev / watch mode so that it doesn't modify unchanged _templ.go files
I love using dev mode; it's a super productive feature! Thanks for adding it 😄

The feature could be improved by ensuring that dev mode does not cause a dirty working copy for files that have not been modified. I find myself perilously close to accidentally committing modified `<file>_templ.go` files, where the diff is basically:

```
-               _, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString("<the contents of the rendered template here>")
+               templ_7745c5c3_Err = templ.WriteWatchModeString(templ_7745c5c3_Buffer, 1)
```

I get why it is written this way, but it would be absolutely **wonderful** if I could run dev mode, make changes to my `.templ` files, and be able to commit modified `_templ.go` files without fear of committing the wrong thing.

This would suggest an alternate mechanism for determining whether the go application is in dev mode at runtime. A few ideas of the top of my head:

* It could be an alternate mode requiring the use of an environment variable (`TEMPL_DEV_MODE`)
* It could use detection of the `_templ.txt` files, and automatically load them if they are present (perhaps _paired_ with an environment variable to avoid this work when running a production application)
* Other options?

a-h:
- **first attempt**
- **fix event handler**
- **fix double quoting**
- **it helps to use the right variable name**
- **stop double quoting**
- **quote again**
- **stop quoting, but add quotes**
- **use devMode from the CLI flags on the CLI, leave env for runtime**
- **fix up watch mode**
- **add generated files**


a-h:
Fixes #700

joerdav:
And will review this Monday too :)

watchmode.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+	_, err = io.WriteString(w, s)
+	return err
+}
+
+// WriteWatchModeString is used when rendering templates in development mode.
+// the generator would have written non-go code to the _templ.txt file, which
+// is then read by this function and written to the output.
+// Deprecated: since templ v0.3.x generated code uses WriteString.
```
joerdav, line 48: Oddly, for linters to detect the `Deprecated` comment, it has to be at the start of a paragraph in the comment (https://go.dev/wiki/Deprecated). So this needs to either be:

```go
// WriteWatchModeString is used when rendering templates in development mode.
// the generator would have written non-go code to the _templ.txt file, which
// is then read by this function and written to the output.
//
// Deprecated: since templ v0.3.x generated code uses WriteString.
```

or

```go
// Deprecated: since templ v0.3.x generated code uses WriteString.
// WriteWatchModeString is used when rendering templates in development mode.
// the generator would have written non-go code to the _templ.txt file, which
// is then read by this function and written to the output.
```

watchmode.go
```diff
@@ -0,0 +1,132 @@
...
+	"os"
+	"runtime"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+var developmentMode = os.Getenv("TEMPL_DEV_MODE") == "true"
+
```
joerdav, line 16: As a new runtime function, do we want to place this in the `runtime` package?

cmd/templ/generatecmd/eventhandler.go
```diff
@@ -80,71 +79,73 @@ type FSEventHandler struct {
...
 	genSourceMapVis            bool
-	DevMode                    bool
 	Errors                     []error
 	keepOrphanedFiles          bool
 	writer                     func(string, []byte) error
 	lazy                       bool
 }
 
-func (h *FSEventHandler) HandleEvent(ctx context.Context, event fsnotify.Event) (goUpdated, textUpdated bool, err error) {
+func (h *FSEventHandler) HandleEvent(ctx context.Context, event fsnotify.Event) (updated, goUpdated, textUpdated bool, err error) {
```
joerdav, line 95: Is it worth this returning a struct now that there is 3 non-error reture values?

joerdav: This should make the returns easier to read:

```go
if h.devMode {
	// Don't delete the file if we're in dev mode, but mark that text was updated.
	result.TextUpdated = true
	return result, nil
}
```
or
```go
if h.devMode {
	// Don't delete the file if we're in dev mode, but mark that text was updated.
	result := EventResult{ TextUpdated: true }
	return result, nil
}
```

cmd/templ/generatecmd/eventhandler.go
```diff
@@ -212,68 +213,78 @@ func (h *FSEventHandler) UpsertHash(fileName string, hash [sha256.Size]byte) (up
 
 // generate Go code for a single template.
 // If a basePath is provided, the filename included in error messages is relative to it.
-func (h *FSEventHandler) generate(ctx context.Context, fileName string) (goUpdated, textUpdated bool, diagnostics []parser.Diagnostic, err error) {
+func (h *FSEventHandler) generate(ctx context.Context, fileName string) (updated, goUpdated, textUpdated bool, diagnostics []parser.Diagnostic, err error) {
```
joerdav, line 216: This function may also benefit from a typed return value, thoughts?

cmd/templ/generatecmd/cmd.go
```diff
@@ -262,7 +264,11 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
 				postGenerationEventsWG.Add(1)
 				if cmd.Args.Command != "" && goUpdated {
 					cmd.Log.Debug("Executing command", slog.String("command", cmd.Args.Command))
-					if _, err := run.Run(ctx, cmd.Args.Path, cmd.Args.Command); err != nil {
+					var env []string
```
joerdav, line 267: It's probably okay to just use `os.Setenv` rather than the additional parameter to the run func. 

In tests you can use `t.Setenv` https://pkg.go.dev/testing#T.Setenv

joerdav (changes requested):
Nice! This'll make developing templ much easier, no more stop-starting! 

I had a couple of comments, nothing major.

watchmode.go
a-h: Good point. I wanted to reuse the existing functions and memory so that using templ 0.2.x generated code and 0.3.x generated code wouldn't cache the values twice, but given the plan to deprecate having runtime stuff in the templ package, I'll make the move.

joerdav (approved):

joerdav:
Nice, all looks great!

Issue #1005, opened 2024-11-25
Excessive CPU and memory usage caused by a wrong syntax (unclosed brackets)


cykuotw:
There is probably an unexpected situation due to unclosed brackets that trigger excessive CPU and memory usage in `templ fmt` and `templ generate`.

I have a templ component defined as 
```go
templ SplitRule(user types.GroupMember, members []types.GroupMember)
```
When I call it in another templ file but accidentally forget to close the brackets like:
```go
@SplitRule(types.GroupMember{
    UserID:   uuid.NewString(),
    Username: "user me",
}, []types.GroupMember{
    {
    UserID:   uuid.NewString(),
    Username: "user 1",
    },
// })  <------ should be here but I accidentally forget
```

There are two cases that can trigger the excessive CPU and memory usage:
1. save the code in VSCode and trigger `templ fmt`
    - resolve: disable/uninstall `templ-vscode` vscode plugin and reload vscode  
2. use `templ generate`
    - resolve: kill terminal process

Both of the **2 latest versions** can reproduce the problem.

Just in case you want to reproduce it, here is the [repo](https://github.com/cykuotw/expense-tracker) and here is the [file](https://github.com/cykuotw/expense-tracker/blob/main/frontend/views/components/newExpense.templ)

**Snapshots**
- `top -i` output,
![image](https://github.com/user-attachments/assets/0ce27ed8-37f4-42b7-bc09-8fdab8097f4f)
- `templ generate -v` output
![image](https://github.com/user-attachments/assets/b9ec39bb-01a7-44c1-8d56-1e3ddf56ee95)

**`templ info` output**
```bash
(✓) os [ goos=linux goarch=amd64 ]
(✓) go [ location=/usr/local/go/bin/go version=go version go1.23.3 linux/amd64 ]
(✓) gopls [ location=/home/cykuotw/go/bin/gopls version=golang.org/x/tools/gopls v0.16.2 ]
(✓) templ [ location=/home/cykuotw/go/bin/templ version=v0.2.793 ]
```

**Desktop:**
 - OS: WSL Ubuntu 20.04.6 LTS
- templ CLI version: v0.2.793
- Go version: go1.23.3 linux/amd64
- `gopls` version: golang.org/x/tools/gopls v0.16.2

a-h:
Thanks for the report. Likely an issue with the parser.

The Go expression parser is here: https://github.com/a-h/templ/tree/main/parser/v2/goexpression
[...]

PR #1029, opened 2024-12-28
fix: parser hangs when a bracket isn't closed (#1005)

harrisbisset:
If a bracket (curly or parentheses) isn't closed, the parser keeps reading until it is closed.
However, the parser NEVER checks if EOF is reached. So, the parser loops when EOF and there's a missing bracket.

## Solution:
Check for EOF before handling anything else (could be elsewhere, but I think this makes the most sense).
If the parser is expecting nested functions, then it should return an error with the unclosed bracket.

#### Templ:
```go
package example

templ Show() {
    @Exists(
}
```

#### Result:
```
(✗) Error [ error=failed to generate code for "path_here/example.templ": /path_here/example.templ parsing error: templ element: invalid go expression: unbalanced '(': line 3, col 5 ]
(✗) Command failed: generation completed with 1 errors
```

Closes https://github.com/a-h/templ/issues/1005


a-h:
Thanks for your contribution, much appreciated.

PR #1031, opened 2024-12-30
chore: bump version of go tools, go compiler, and goquery (fixes CVE)

a-h:
See https://github.com/advisories/GHSA-w32m-9786-jp63


a-h:
Unfortunately, goquery has bumped to Go 1.23 to support iterators, as well as fixing a CVE which affects the CLI, where goquery is a dependency.

Usually, with templ, I like to stay on the oldest supported version of Go, but here we're moving from Go language version 1.21  to 1.23, which will mean that users will need to upgrade to Go 1.23.

Go 1.23 was released 4 months ago, so it's not completely bleeding edge. Many users may have upgraded to Go 1.23.

Since this release contains a breaking change to CLI workflows, I think now is a good time to bump the Go version too.

PR #1037, opened 2025-01-03
feat: imports grouping

### LINKED ISSUE #939: Allow import grouping
Usually when I write code I like to organize imports by category: std, a project im working on, other packages. In golang it's possible to put a new line between imports and they they will be sorted separately. Templ formatter removes all spaces and sorts them as one block

MilkeeyCat:
This pr changes formatter to keep a new line between imports. For example, if you had such code:
```go
package main

import (
	"fmt"
	"strings"

	"strconv"
)

var _, _ = fmt.Print("")
var _ = strings.Index("", "")
var _ = strconv.Quote("")
```

It would remove the empty line, but now it doesn't.

closes #939

### DIFF
--- a/cmd/templ/imports/process.go
+++ b/cmd/templ/imports/process.go
@@ -104,21 +104,33 @@ func Process(t parser.TemplateFile) (parser.TemplateFile, error) {
 	if err := eg.Wait(); err != nil {
 		return t, err
 	}
-
-	// Delete all the existing imports.
+	// Delete unused imports.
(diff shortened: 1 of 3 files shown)


a-h:
Yes, but if there's any other behaviour to consider, like what happens if there's two line breaks, something to cover those scenarios would also be helpful.

Thanks!

a-h (approved):

a-h:
Thanks for that, much appreciated!

PR #1048, opened 2025-01-14
docs: clarify when live reload might fail due to middleware implementation

zachczx:
Hello, I didn't realize I haven't been benefiting from live reload despite setting it up in my Makefile. After digging about today it turns out my use of https://github.com/go-swiss/compress Brotli middleware was causing issues, and that is not the case when using other gzip or middleware implementation. Without diving deeper, it seems weird that this compression middleware creates a problem, because it is using the same Brotli package as templ's under the hood for proxy.go. Probably a case of implementation where compressed handler got written to ResponseWriter and flushed, so there was nothing to inspect when templ got its chance.

Regardless, I think it'd be a good idea to clarify upfront that the Content-Type is important for the JS script insertion and hint at middleware implementation being an issue.

### DIFF
--- a/docs/docs/09-developer-tools/03-live-reload.md
+++ b/docs/docs/09-developer-tools/03-live-reload.md
@@ -6,6 +6,7 @@ To access a Go web application that uses templ in a web browser, a few things mu
 2. The Go code must start a web server on a port, e.g. (`http.ListenAndServe("localhost:8080", nil)`.
 3. The Go program must be ran, e.g. by running `go run .`.
 4. The web browser must access or reload the page, e.g. `http://localhost:8080`.
+5. Content-Type must be text/html. 
 
 If the `*.templ` files change, #1 and #2 must be ran.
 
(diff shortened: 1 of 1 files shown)


a-h:
I'm interested in why this _doesn't_ work. There's tests for brotli in https://github.com/a-h/templ/blob/725a88f73cfbaff9e43fbf50b6abbae59630bfa9/cmd/templ/generatecmd/proxy/proxy_test.go#L348

a-h (approved):

a-h:
Thanks, I had a look at the Brotli middleware and couldn't see where that issue might be. But... also, there was an issue with the watch mode that I just resolved (not released yet), so it might have been that. 🤦🏻

zachczx:
> Thanks, I had a look at the Brotli middleware and couldn't see where that issue might be. But... also, there was an issue with the watch mode that I just resolved (not released yet), so it might have been that. 🤦🏻

I'll open another issue if I have the time to dig deeper, regardless I appreciate the time you took to look into this!

PR #1053, opened 2025-01-22
refactor: don't generate a `type` attribute for `<script>` elements, update associated docs

therealparmesh:
[From the MDN docs about the script element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attribute_is_not_set_default_an_empty_string_or_a_javascript_mime_type):

> Attribute is not set (default), an empty string, or a JavaScript MIME type
> 
> Indicates that the script is a "classic script", containing JavaScript code. Authors are encouraged to omit the attribute if the script refers to JavaScript code rather than specify a MIME type. JavaScript MIME types are listed in the IANA media types specification.

I think it'd be good to reflect the best practices for `<script>` tag usage in templ docs.

Thank you for this library, and keep up the great work!

### DIFF
--- a/docs/docs/03-syntax-and-usage/03-attributes.md
+++ b/docs/docs/03-syntax-and-usage/03-attributes.md
@@ -225,7 +225,7 @@ templ Button(text string) {
 ```
 
 ```html title="Output"
-<script type="text/javascript">
+<script>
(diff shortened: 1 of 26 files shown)


a-h (approved):
Great, thanks a lot.

PR #1056, opened 2025-01-28
Add `//go:generate` directive to generated files

utilyre:
This PR adds `//go:generate` directive to all generated files, enabling devs to re-generate that file with the `go generate` command.

The aim is to make it easier and more portable (in docker containers, etc) to re-generate `*_templ.go` files.

Appreciate your work btw :)


joerdav:
I'm not sure I fully am on board with this feature just yet. What if the command used to generate is different from the default?

If the `-include-version=false` flag is provided for example, a subsequent `go generate` would undo that.

As a heads up we do have a bit of a contribution guide here: https://github.com/a-h/templ/blob/main/CONTRIBUTING.md#come-up-with-a-design-and-share-it

Mostly so that we can talk through ideas before they are implemented:

> Before starting work on any major pull requests or code changes, start a discussion at https://github.com/a-h/templ/discussions or raise an issue.
> 
> We don't want you to spend time on a PR or feature that ultimately doesn't get merged because it doesn't fit with the project goals, or the design doesn't work for some reason.

utilyre:
> What if the command used to generate is different from the default?

I see, that would be an issue.

From what I've found, these are the options that can affect the output:

- `-include-version`
- `-include-timestamp`
- and potentially `-cmd`

@joerdav, can you please confirm that this list is exhaustive?

> As a heads up we do have a bit of a contribution guide here: https://github.com/a-h/templ/blob/main/CONTRIBUTING.md#come-up-with-a-design-and-share-it
>
> Mostly so that we can talk through ideas before they are implemented:
>
> > Before starting work on any major pull requests or code changes, start a discussion at https://github.com/a-h/templ/discussions or raise an issue.
> >
> > We don't want you to spend time on a PR or feature that ultimately doesn't get merged because it doesn't fit with the project goals, or the design doesn't work for some reason.

Thanks for the heads up. I didn't think this was a major PR, that's why I didn't open any issues or discussions beforehand.

joerdav:
It's hard to quantify "major PR" I suppose. This one isn't minor from my perspective because it adds something to every generated template. In issues that propose a change we can discuss the impact and make sure we understand it before putting time to implementation.

PR #1070, opened 2025-02-17
refactor: remove goquery dependency

a-h:
goquery is great, but actually, we only need one or two matchers, so there's no need to bring it in.

I've left it in use within an example that shows how to use it in a test.

Overall, am attempting to reduce the number of dependencies where sensible.


htmlfind/htmlfind.go
```diff
@@ -0,0 +1,67 @@
+package htmlfind
```
joerdav, line 1: Should we consider putting this in an `internal` subdirectory?

a-h: Agree.

PR #1073, opened 2025-02-19
Fix data race with errs []error pattern

cristaloleg:


generator/htmldiff/diff.go
```diff
@@ -18,15 +18,15 @@ func DiffStrings(expected, actual string) (diff string, err error) {
 	var wg sync.WaitGroup
 	wg.Add(2)
 
-	var errs []error
+	var errs [2]error
```
a-h, line 21: Thanks for spotting this issue and raising the PR.

Is it worth doing `errs := make([]error, 2)` instead? The array gets turned into a slice anyway later on.

cristaloleg: Yep, can be. Fixed.

joerdav: I think using an array as the backing for a slice can sometimes avoid an allocation, if I remember correctly.

cristaloleg: Slices upto 8k (iirc) are allocated on stack. But let's be honest: this is not a case where performance will be degraded by 1 allocation.

a-h (approved):

Issue #1067, opened 2025-02-12
docs: Document usage of new go tools feature in go1.24.0


joerdav:
With the release of go1.24.0, there is a new feature allowing tools to be tracked in the go.mod.

You can run `go get -tool github.com/a-h/templ/cmd/templ@latest` to add the latest templ version as a tool.

Then you can run templ as normal using `go tool templ`. For example `go tool templ generate`.

This links the generator version to the runtime version specified in the go.mod. So you reduce the likelihood of generated code that is incompatible with the runtime version.

There should be a docs section outlining this, or at least that links to the go docs as a recommended way of tracking templ versions.

PR #1078, opened 2025-02-23
docs: go tool installation and info panel

Quarkz14:
Update templ installation docs to using the new tool directive in Go 1.24 with an info panel linking to the tool directive docs as described here https://github.com/a-h/templ/issues/1067


a-h:
Yes, I agree. You can install templ globally, or you can install it for just your project - like the difference between `npm install` and `npm install -g`.

PR #1085, opened 2025-03-02
docs: improve vim documentation

iefserge:
Update Vim documentation and add LSP plugin config example using `vim-lsp`.
It supports auto-formatting on save using LSP as well.

Also I didn't like how indentation worked in templ files, so I forked [joerdav/templ.vim](https://github.com/Joe-Davidson1802/templ.vim) and made it work more like Go files, but with some additions for opening/closing html tags. I added optional section about it, but can remove if you don't like that.


a-h (approved):

a-h:
Thanks for this, very helpful.

PR #1099, opened 2025-03-15
fix: use temp directory for watch mode files, fixes 1093

a-h:


a-h:
@joerdav - I remember we talked about doing this a while ago, but I couldn't find any issues or anything.

joerdav:
Nice, yeah this will be really good for the developer experience! 

It might be good if we could write some tests for `GetDevModeTextFileName` in a way that works cross platform. It might take some interesting work arounds but would be good to have it covered.

PR #1109, opened 2025-03-23
Document templ.EscapeString a little bit more

MaxLap:
Right now, that function's name is hidden in a single place in the documentation, and searching for "escape" or "escaping" doesn't make it easy to find.

By mentioning it in some relevant places, we make it easier for devs to find it.

### DIFF
--- a/docs/docs/03-syntax-and-usage/04-expressions.md
+++ b/docs/docs/03-syntax-and-usage/04-expressions.md
@@ -101,3 +101,5 @@ templ component() {
 ```html title="Output"
 <div>&lt;/div&gt;&lt;script&gt;alert(&#39;hello!&#39;)&lt;/script&gt;&lt;div&gt;</div>
 ```
+
+Manual escaping can be done with the `templ.EscapeString` function, but then you will need to use `@templ.Raw` to avoid double-escaping. See [Rendering raw HTML](/syntax-and-usage/rendering-raw-html) for details.
--- a/docs/docs/03-syntax-and-usage/16-rendering-raw-html.md
+++ b/docs/docs/03-syntax-and-usage/16-rendering-raw-html.md
@@ -3,7 +3,7 @@
 To render HTML that has come from a trusted source, bypassing all HTML escaping and security mechanisms that templ includes, use the `templ.Raw` function.
 
 :::info
-Only include HTML that comes from a trusted source.
+Only include HTML that comes from a trusted source. Use the `templ.EscapeString` function to escape content that comes from untrusted sources.
 :::
 
 :::warning


docs/docs/03-syntax-and-usage/04-expressions.md
```diff
@@ -101,3 +101,5 @@ templ component() {
 ```html title="Output"
 <div>&lt;/div&gt;&lt;script&gt;alert(&#39;hello!&#39;)&lt;/script&gt;&lt;div&gt;</div>
 ```
+
+Manual escaping can be done with the `templ.EscapeString` function, but then you will need to use `@templ.Raw` to avoid double-escaping. See [Rendering raw HTML](/syntax-and-usage/rendering-raw-html) for details.
```
a-h, line 105: I don't really understand why you've added this.

You don't need to (and shouldn't) manually escape strings, it's done automatically by templ.

docs/docs/03-syntax-and-usage/16-rendering-raw-html.md
```diff
@@ -3,7 +3,7 @@
 To render HTML that has come from a trusted source, bypassing all HTML escaping and security mechanisms that templ includes, use the `templ.Raw` function.
 
 :::info
-Only include HTML that comes from a trusted source.
+Only include HTML that comes from a trusted source. Use the `templ.EscapeString` function to escape content that comes from untrusted sources.
```
a-h, line 6: I'm not really sure why you would want to use the `Raw` component, but then escape the content, when you could use a string expression `{ stringContent }` directly, where the content would automatically be HTLM escaped.

a-h (commented):
Hi, thanks for the PR!

Just got a couple of questions about what you're trying to do, because I don't really understand it, and want to make sure the docs don't confuse people about the escaping behaviour.

If you've got a bit of background on what problems you were facing with templ, and why these docs would have helped, that might help me understand.

Thanks!

a-h:
Out of interest, how come you don't make the changes on the `full_description` instead?

https://go.dev/play/p/jPaUdgoIB4Y

In templ, you'd then be able to use:

```templ
{ strings.Replace(full_description, "_", "_\u00AD", -1) }
```

Is it because you're not using UTF-8 encoding for your HTML, or some other reason?

`templ.EscapeString` is just a wrapper around `html.EscapeString` from the stdlib. The only reason it's in the templ package is so that the code can safely assume that the `import` statement for `templ` will be present.

MaxLap:
I altered the snippet to shorten it. `full_description` is actually a field on a struct that is used elsewhere in the program in contexts where I do not want HTML escaping.

It's good to know about `html.EscapeString`. This is another thing that might be helpful to mention then.

Right now, searching for "EscapeString" only find it in the `Code-only components` section of components. So unless you are lucky about your other searches and land on it. It's basically hidden because that's not where someone expects to have to go to get the answer.

Issue #740, opened 2024-05-17
proposal: deprecate CSS components in favour of 3rd party CSS pre-processors


a-h:
`templ css` CSS components are very basic. They don't support a range of use cases #262 including media queries, and the interpolation features are minimal. I have built a PoC of interpolating CSS values within a standard `<style>` element at https://github.com/a-h/templ-css but before jumping to implementation, I took a step back.

## People are already using pre-processors

Many users of templ are using Tailwind, or other CSS pre-processors already. They are essentially templating libraries for CSS, but can also do thing like include browser specific properties.

## Pre-processors are popular and mature

The PostCSS NPM package gets around 69.3 million downloads per week, while SASS gets around 13.5 million per week, and tailwind gets 8M. They're well established projects, with large user bases.

## templ css features are covered by existing pre-processors

`templ css` was designed so that it was possible to include classes for specific templ components, and only render the CSS if required. It dynamically generates class names to provide basic scoping. But... its design has severe limits.

Given the range of available pre-processors, their popularity, and the limited resources of the templ project, I think it might be better to spend time on improving the DX around using existing pre-processors in templ, rather than to attempt to create one, alongside everything else.

PostCSS has a plugin for scoping at https://github.com/madyankin/postcss-modules - which allows the scoping behaviour to be achieved easily, and there's a plugin for PostCSS which outputs a list of classes - https://www.npmjs.com/package/postcss-ts-classnames - in that particular case, in TypeScript format.

## We can get compile-time errors about missing CSS classes

If, for example, we created a version of `postcss-ts-classnames` PostCSS plugin which created a Go package called `class` containing a load of constants, we'd then be able to import the package into our templates, and get a strongly defined list of classes. When you write `<div class={ class.` in templ, the Go LSP would provide a list of classes in the IDE autocomplete.

This would give us the benefits of a strongly typed list of classes, for minimal development effort.

## This proposal fits well with #739

For JavaScript handling, I've proposed an alternative focus on to using JSON `script` elements, alongside `esbuild` to bundle up scripts. It solves the problems of transpiling TypeScript so you can use async/await, sorts out minification etc. you can see what that looks like at https://github.com/a-h/templ/issues/739 - esbuild is written in Go, and is very fast.

## `templ.OncePerRequest`

To ensure that links to stylesheets, or `<style>` content itself is only rendered once, a new `templ.OncePerRequest` component could ensure that children of the component are only rendered once per request, using the `ctx` parameter.

## Build process

[...]

PR #1114, opened 2025-03-29
feat: allow the AST to be modified using a visitor

matthewmueller:
This PR:
- Adjusts the AST to use pointers, so it can be modified in place
- Implements a `Visitor` interface that allows the AST to be traversed and rewritten

The goal of this PR is to allow the community to create compile-time plugins for Templ. Some examples could include:
- Scoped CSS in https://github.com/a-h/templ/issues/740
- HTML minification

We could also generalize a few patterns:
- Generating Go code could be a visitor
- Formatting with imports could be another visitor

This PR ended up a bit larger than I would have liked. While you could implement the `Visitor` pattern without making the AST pointers, it effectively makes it so the AST is read-only. Being able to modify the AST opens up a lot more possibilities.

Let me know if this is a direction you're interested in heading. Tests pass, but there's a few more tests I'd like to contribute before merging.

Also in the #templ channel in the Gophers Slack if you'd like to chat further first.


a-h (approved):

matthewmueller:
Thanks!

PR #1118, opened 2025-03-31
Conditionalize language server preload

### LINKED ISSUE #1117: Conditionalize language server preload behavior
The Templ server currently preloads all Templ files within the workspace. This causes performance issues in large monorepos where thousands of Templ files exist.

Include a command line flag to turn this behavior off while keeping it on by default to avoid a breaking change.

prathshenoy:
Allow disabling of language server preload through a command line flag to fix performance issues in large monorepos.

Closes #1117.

### DIFF
--- a/cmd/templ/lspcmd/proxy/server.go
+++ b/cmd/templ/lspcmd/proxy/server.go
@@ -240,7 +240,7 @@ func (p *Server) Initialize(ctx context.Context, params *lsp.InitializeParams) (
 	}
 
 	if !p.NoPreload {
-		p.preload(params.WorkspaceFolders)
+		p.preload(ctx, params.WorkspaceFolders)
 	}
 
 	result.ServerInfo.Name = "templ-lsp"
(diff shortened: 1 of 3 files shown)


cmd/templ/lspcmd/proxy/server.go
```diff
@@ -237,58 +239,60 @@ func (p *Server) Initialize(ctx context.Context, params *lsp.InitializeParams) (
...
-				},
-			}
-
-			p.preLoadURIs = append(p.preLoadURIs, didOpenParams)
-			return nil
-		})
-		if werr != nil {
-			p.Log.Error("walk error", slog.Any("error", werr))
 		}
 	}
```
JamyDev, line 297: I think it may be worthwhile to move the entire preload into it's own method:

```
func (p *Server) preloadWorkspaceFolders(workspaceFolders []lsp.WorkspaceFolder) {
   for ...
}
```
```suggestion
	p.preloadWorkspaceFolders(params.WorkspaceFolders)
```

cmd/templ/main.go
```diff
@@ -359,6 +359,8 @@ Args:
     Enable pprof web server (default address is localhost:9999)
   -http string
     Enable http debug server by setting a listen address (e.g. localhost:7474)
+  -noPreload
```
JamyDev, line 362:
```suggestion
  -no-preload
```

Follow convention

prathshenoy: I don't think `no-preload` would follow convention since the other flags above use camel case.

joerdav:
Happy to approve subject to my comment on the flag naming! Thanks again for the contribution.

cmd/templ/main.go
prathshenoy: I see. Thanks for the correction. Fixed!

JamyDev (approved):

joerdav (approved):

prathshenoy:
Oops, I should've rebuilt that change. Fixed!

joerdav:
Great! This will be in the next release. Thanks again!

Issue #1121, opened 2025-04-08
LSP diagnostics not working on Windows


koppa96:
**Describe the bug**
LSP diagnostic errors don't show up in neither VS Code, nor in Goland.

**To Reproduce**
https://github.com/koppa96/templ-lsp-repro

**Expected behavior**
I expect an error to be visible when I use a variable that is not declared in a scope.

**Screenshots**

![Image](https://github.com/user-attachments/assets/6e3bb95f-18e4-4940-90b1-b56e67d09527)

The variable ajwskerf is clearly not declared, but is not underlined, and the IDE shows no problems about the file. Same thing happens in VS Code.

**Logs**

```
{"time":"2025-04-08T13:29:26.1674588+02:00","level":"INFO","msg":"lsp: starting up..."}
{"time":"2025-04-08T13:29:26.173959+02:00","level":"INFO","msg":"lsp: starting gopls..."}
{"time":"2025-04-08T13:29:26.1894604+02:00","level":"INFO","msg":"creating gopls client"}
{"time":"2025-04-08T13:29:26.189963+02:00","level":"INFO","msg":"creating proxy"}
{"time":"2025-04-08T13:29:26.189963+02:00","level":"INFO","msg":"creating templ server"}
{"time":"2025-04-08T13:29:26.189963+02:00","level":"INFO","msg":"listening"}
{"time":"2025-04-08T13:29:26.1909605+02:00","level":"INFO","msg":"client -> server: Initialize"}
{"time":"2025-04-08T13:29:26.3146584+02:00","level":"ERROR","msg":"walk error","error":"CreateFile /c%3A/Users/koppa.peter/GolandProjects/templtest: A rendszer nem találja a megadott elérési utat."}
{"time":"2025-04-08T13:29:26.3162292+02:00","level":"INFO","msg":"client -> server: Initialize end"}
{"time":"2025-04-08T13:29:26.3298162+02:00","level":"INFO","msg":"client -> server: Initialized"}
{"time":"2025-04-08T13:29:26.3298162+02:00","level":"INFO","msg":"client -> server: Initialized end"}
{"time":"2025-04-08T13:29:26.3303164+02:00","level":"INFO","msg":"client <- server: ShowMessage","message":"Loading packages..."}
{"time":"2025-04-08T13:29:26.5195718+02:00","level":"INFO","msg":"client -> server: DidOpen","uri":"file:///c%3A/Users/koppa.peter/GolandProjects/templtest/hello.templ"}
[...]

PR #1122, opened 2025-04-08
fix: run multiplatform tests

joerdav:
@a-h I thought we could do with regularly running the tests against windows machines!


joerdav:
@a-h The PR fails, but not because of the PR. Do we fix the issues with windows in this PR or merge this in and fix them separately?

a-h:
I would probably prefer to fix windows issues in the PR if possible.

I'll take a look at what the issues are when I get chance.

joerdav:
Fair, suppose it doesn't make sense to break main.

joerdav:
I think this is related:

https://github.com/a-h/templ/issues/1121#issuecomment-2790643875

Issue #1135, opened 2025-04-25
`X expected nodes, but none were found` when writing _hyperscript in a `<script>` block


zyriab:
**Describe the bug**
Since the last update (no issue with the previous [v0.3.833] version); when writing [_hyperscript](https://hyperscript.org/)'s possessive `'s` commands in a `<script>` tag, it breaks Templ LSP.
My guess is that the `'` is detected as a string opening apostrophe by the LSP.

The exact error message I have is:
```
credits-modal.templ parsing error: @handler.Once(): expected nodes, but none were found: line 354, col 0 ]
```

**To Reproduce**
Here's the script I have in production that now breaks the generation:

```applescript
templ Script() {
	{{ handler := templ.NewOnceHandle() }}
	@handler.Once() {
	<script type="text/hyperscript">
            def calculate_price(count)
                set tier_1 to #tier-1's value
                set tier_2 to #tier-2's value
                set tier_3 to #tier-3's value

                set tier_1_price to (tier_1.split("=")[1]) as an Int
                set tier_2_count to (tier_2.split("=")[0]) as an Int
                set tier_2_price to (tier_2.split("=")[1]) as an Int
                set tier_3_count to (tier_3.split("=")[0]) as an Int
                set tier_3_price to (tier_3.split("=")[1]) as an Int

                if count is greater than or equal to tier_3_count 
                    return count * (tier_3_price/100)
                end

                if count is greater than or equal to tier_2_count
                    return count * (tier_2_price/100)
                end

                return count * (tier_1_price/100)
            end
        </script>
	}
}
```

Templ generation works if I remove the 3 lines with the `'s`.

**Expected behavior**
It should not error, as it used to do before I updated from 0.3.833 to 0.3.857.

**`templ info` output**
```
(✓) os [ goos=darwin goarch=amd64 ]
(✓) go [ location=/usr/local/go/bin/go version=go version go1.24.2 darwin/amd64 ]
(✓) gopls [ location=/Users/zyr/go/bin/gopls version=golang.org/x/tools/gopls v0.16.1 ]
(✓) templ [ location=/Users/zyr/go/bin/templ version=v0.3.857 ]
```

**Desktop (please complete the following information):**
 - OS: MacOS 15.4.1 (Intel i7)
 - templ CLI version 0.3.857 (used to work on 0.3.833)
- Go version: 1.24.2
- gopls version: 0.18.1

**Additional context**
I understand if you don't want to fix this as we must be very few using this quite obscure language 😅 

[...]

PR #1136, opened 2025-04-25
fix: ignore non js scripting languages in script parser

joerdav:
Fixes https://github.com/a-h/templ/issues/1135

### DIFF
--- a/parser/v2/scriptparser.go
+++ b/parser/v2/scriptparser.go
@@ -1,6 +1,7 @@
 package parser
 
 import (
+	"slices"
 	"strings"
 
 	"github.com/a-h/parse"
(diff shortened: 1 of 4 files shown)


parser/v2/scriptparser.go
```diff
@@ -58,6 +59,21 @@ func (p scriptElementParser) Parse(pi *parse.Input) (n Node, ok bool, err error)
 		return
 	}
 
+	// Check for a type attribute. If it is not text/javascript, we need to parse the contents as raw text.
```
a-h, line 62: Is this comment correct?

The behaviour should be that "If there's a type attribute _and_ it's not text/javascript, then we need to parse the contents as raw text.", since, without a type attribute, the default is JavaScript.

parser/v2/scriptparser.go
```diff
@@ -152,14 +168,25 @@ loop:
 	return e, true, nil
 }
 
-var jsEndTag = parse.String("</script>")
-var endTagStart = parse.String("</")
+func isNotJavascript(a Attribute) bool {
```
a-h, line 171: This looks like the correct behaviour.

a-h (commented):
LGTM apart from the comment. If you can confirm the expected behaviour, that would be great.

joerdav:
Makes sense, updated the comment!

PR #1139, opened 2025-04-27
feat: support for ordered attributes

### LINKED ISSUE #1138: Preserve order of `data-` attributes when using `templ.Attributes`
When passing `templ.Attributes` to a component (templUI in this case), the attributes appear to be added in alphabetical order, rather than insert order.

For example:

```templ
@alert.Alert(alert.Props{ID: id, Attributes: templ.Attributes{
	"data-signals": "{showAlert: true}",
	"data-show":    "$showAlert",
}}) {Some alert text.}
```

Will end up with `data-show` being added before `data-signals` (which causes an error in my app).

According to https://html.spec.whatwg.org/#embedding-custom-non-visible-data-with-the-data-*-attributes , 

> To get a DOMStringMap's name-value pairs, run the following algorithm:
> 
> Let list be an empty list of name-value pairs.
> 
> For each content attribute on the [DOMStringMap](https://html.spec.whatwg.org/#domstringmap)'s  [associated element](https://html.spec.whatwg.org/#concept-domstringmap-element) whose first five characters are the string "data-" and whose remaining characters (if any) do not include any [ASCII upper alphas](https://infra.spec.whatwg.org/#ascii-upper-alpha), **in the order that those attributes are listed in the element's [attribute list](https://dom.spec.whatwg.org/#concept-element-attribute)**, add a name-value pair to list whose name is the attribute's name with the first five characters removed and whose value is the  attribute's value.

The key bit being that the attributes are available to JS "in the order that those attributes are listed in the element's attribute list".

(Appreciate that this one is a tricky one, given that Attributes is a map under the covers.)

garrettladley:
resolves #1138

adds the ability to have ordered attributes.

the difficulty comes from the fact that `templ.Attributes` is a map under the hood.

added a new type, `templ.OrderedAttributes` for the use case where order of insertion matters while not modifying the external behavior of `templ.Attributes`


garrettladley:
not sure why build is failing say it doesn't have the package. i added the dependency with `go get -u github.com/wk8/go-ordered-map/v2` and the diff to go.mod and go.sum reflects this

a-h:
Brilliant, would you be able to do this without taking on any additional 3rd party dependencies? I'm trying to keep close-ish to stdlib to reduce supply chain concerns.

I wrote an example of an ordered map you can copy/paste into templ: https://github.com/a-h/orderedmap - or feel free to make your own implementation that better suits the use case etc.

a-h:
The build fails because I use Nix to build the output binaries, which has its own hashing system of dependencies, because of the way that Go creates hashes in go.mod. I can sort that out, it's just a case of updating the flake.nix hash value, but if no dependencies change, then the build won't fail.

delaneyj:
Just a heads up, per the [HTML spec](https://html.spec.whatwg.org/#custom-data-attribute) at least all data-* attributes must be a [StringMap](https://html.spec.whatwg.org/#domstringmap), which is insertion order always.  This order :100: matters for tools like Datastar.  Insertion order should not be a flag, it should match the spec if possible.

garrettladley:
@a-h slept on it and was able to simplify the implementation to not even require an ordered map implementation 😁

garrettladley:
@delaneyj i agree that matching the HTML spec is essential. however, changing the signature of `templ.Attributes` would be a breaking change. i think introducing and using `templ.OrderedAttributes` is a simple but effective way to resolve #1138 while also not breaking for others.

curious to hear @a-h's thoughts on potentially making `templ.Attributes` equivalent to `templ.OrderedAttributes` in a future version of templ

a-h:
Would be good to normalise everything to an order-preserving map, but priority is to not break people's code. I expect it's a very small proportion of users that would be affected, but still... 😁

garrettladley:
so you think this PR is good to go?
with an order-preserving map to come later down the line

delaneyj:
> @delaneyj i agree that matching the HTML spec is essential. however, changing the signature of `templ.Attributes` would be a breaking change. i think introducing and using `templ.OrderedAttributes` is a simple but effective way to resolve #1138 while also not breaking for others.
> 
> curious to hear @a-h's thoughts on potentially making `templ.Attributes` equivalent to `templ.OrderedAttributes` in a future version of templ


You might need to just disagree on this. I think if you rely on invalid behavior then the onerous is on you to match the spec

garrettladley:
@delaneyj i hear you. want me to make this PR the breaking change for `templ.Attributes` to match the spec?

delaneyj:
I think matching the spec is the long term correct solution but leave it to @joerdav and @.  I will say this already 💯 effects Datastar currently and is not a hypothetical issue

delaneyj:
#1132 does help alleviate most of the real world issues for me.  I think this is a solid plan @joerdav

a-h (approved):

PR #1144, opened 2025-05-06
fix: include all watch patterns in consideration for cmd restart

acaloiaro:
When `-watch-pattern` is set and a non-go, non-text (`_templ.txt`) file changes that matches the pattern, processing the change is effectively a no-op because templ only restarts the `-cmd` when Go files change.

A use case for this change is when one uses `//go:embed` directives to embed non-Go assets in go binaries. E.g. when changes are made to `app.js` and `-cmd` is set to `go run main.go`, templ must restart the `-cmd` for changes to be seen. 
 
This change causes templ to re-run its "cmd" argument any time files matching its -watch-pattern change.

### DIFF
--- a/cmd/templ/generatecmd/cmd.go
+++ b/cmd/templ/generatecmd/cmd.go
@@ -57,10 +57,10 @@ type Generate struct {
 }
 
 type GenerationEvent struct {
-	Event       fsnotify.Event
-	Updated     bool
-	GoUpdated   bool
-	TextUpdated bool
+	Event            fsnotify.Event
+	WatchfileUpdated bool
+	GoUpdated        bool
+	TextUpdated      bool
 }
 
 func (cmd Generate) Run(ctx context.Context) (err error) {
@@ -229,15 +229,15 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
 				if err != nil {
 					errs <- err
 				}
-				if !(r.GoUpdated || r.TextUpdated) {
+				if !(r.GoUpdated || r.TextUpdated || r.WatchfileUpdated) {
 					cmd.Log.Debug("File not updated", slog.String("file", event.Name))
 					return
 				}
 				e := &GenerationEvent{
-					Event:       event,
-					Updated:     r.Updated,
-					GoUpdated:   r.GoUpdated,
-					TextUpdated: r.TextUpdated,
+					Event:            event,
+					WatchfileUpdated: r.WatchfileUpdated,
+					GoUpdated:        r.GoUpdated,
+					TextUpdated:      r.TextUpdated,
 				}
 				cmd.Log.Debug("File updated", slog.String("file", event.Name))
 				postGeneration <- e
@@ -256,7 +256,7 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
 		defer postGenerationWG.Done()
 		cmd.Log.Debug("Starting post-generation handler")
 		timeout := time.NewTimer(time.Hour * 24 * 365)
-		var goUpdated, textUpdated bool
+		var goUpdated, textUpdated, watchfileUpdated bool
 		var p *proxy.Handler
 		for {
 			select {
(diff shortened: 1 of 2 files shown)


acaloiaro:
Fixes #1145

acaloiaro:
It's worth noting that up until now, nothing read the value of `GenerationEvent.Updated`. It was written, but not used to support any templ features.

a-h:
Thanks! I've had a read through, but think we need to clarify expected behaviour and document it.

Overall, the docs on live reload weren't very good because of various updates to behaviour over time, so I've updated them (see https://github.com/a-h/templ/blob/main/docs/docs/09-developer-tools/03-live-reload.md)

I think the desired behaviour is that...

- The `--watch-pattern` argument determines which files trigger any kind of action from the `templ --watch` command.
  - This should now default to `*.templ` files and `*.go` files, because `_templ.txt` files are now created in a tmp dir, not in the project dir using the new `TEMPL_DEV_MODE` feature.
- In the case of a change to a `*.templ` file:
  - The file is generated.
  - Generation produces a set of text files and Go code.
  - If only the text has changed, the change is classed as a `TextUpdated` change.
  - If any of the Go expressions in the template have changed, it's classed as a `GoUpdated` change
- In the case of a change to a `*.go` file:
  - If the file is `*_templ.go`, it is ignored, because this is handled as part of `*.templ` file handling.
  - If the file has changed, it's classed as a `GoUpdated` change. 
- In the case of any other file that matches the `--watch-pattern`:
  - It is classed as a `WatchedFileUpdated` change.

At the end of the cycle:

- If any of the changes are `GoUpdated` or `WatchedFileUpdated`, the cmd needs to be restarted.
- If a `GoUpdated`, `TextUpdated` or `WatchedFileUpdated` event has happened, then a reload event needs to be fired.

Is that your expectation?

acaloiaro:
@a-h That's an an accurate summary of the expected behavior. 

Now that you've pointed it out, I think this change set could mention `--watch-pattern` in https://github.com/a-h/templ/blob/main/docs/docs/09-developer-tools/03-live-reload.md). 

This line

> The `templ generate --watch` argument tells templ to watch for changes to `*.templ` and `*.go` files in the current directory.

could read something like 

> The `templ generate --watch` argument tells templ to watch for changes to `*.templ` and `*.go` files in the current directory. By default, `*.go` and `*.templ` files are watched, which is controlled with the `--watch-pattern` switch. 

Without this mention, some users may reach for `wgo`, `air`, or `reflex` prematurely. IMO `--watch` is a pretty decent replacement for all of these tools. It doesn't cover all their use cases or range of complexity, but that's a good thing. 

Do you think this live reload documentation is worth adding in this change set, or do you want to keep that document simpler?

cmd/templ/generatecmd/cmd.go
```diff
@@ -57,10 +57,10 @@ type Generate struct {
 }
 
 type GenerationEvent struct {
-	Event       fsnotify.Event
-	Updated     bool
-	GoUpdated   bool
-	TextUpdated bool
+	Event            fsnotify.Event
+	WatchfileUpdated bool
```
acaloiaro, line 61: @a-h documented this property as `WatchedFileUpdated`. Let's use that name.

cmd/templ/generatecmd/cmd.go
```diff
@@ -266,8 +266,9 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
 					return
 				}
 				goUpdated = goUpdated || ge.GoUpdated
+				watchfileUpdated = watchfileUpdated || ge.WatchfileUpdated
```
acaloiaro, line 269: Refactor: `watchFileUpdated` -> `watchedFileUpdated`

cmd/templ/generatecmd/eventhandler.go
```diff
@@ -94,8 +94,8 @@ type FSEventHandler struct {
 }
 
 type GenerateResult struct {
-	// Updated indicates that the file was updated.
-	Updated bool
+	// WatchfileUpdated indicates that a file matching the watch pattern was updated.
+	WatchfileUpdated bool
```
acaloiaro, line 98: Refactor: WatchFileUpdated -> WatchedFileUpdated

acaloiaro (commented):
Changes from the change set discussion

a-h:
> I think that usage of GoUpdated is equivalent to WatchedFileUpdated, but I'll let someone with more experience with templ's internals weigh in.

Yes, I think you're right. I'll take a look at the code with fresh eyes. There are two actions:

- Restart the program
- Reload the browser

And the logic:

- Watched file has changed:
  - Not a templ file - restart and reload.
  - Is a templ file - restart if go changed, reload.

PR #1147, opened 2025-05-06
feat: gracefully shutdown on sigterm in `-watch` mode 

acaloiaro:
To handle graceful shutdown requests, templ needs to handle `SIGTERM` signals from the operating system.

Templ is difficult to use in signal-based systems like `process-compose` because it does not gracefully shutdown, and thus, is unable to gracefully restart when externally signalled to do so.


a-h (approved):

a-h:
Thanks!

PR #1155, opened 2025-05-20
feat: allow partial parsing for improved LSP support

### LINKED ISSUE #1086: bug(lsp): autocomplete missing on branching statements
**Describe the bug**
I don't get any auto completion on .templ files for branching statements such as for loops and ifs. I do get autocomplete on expressions though.

**To Reproduce**
Try to get auto complete on neovim on branching statements.
I don't know if this is helpful, but here is my config: https://github.com/phenpessoa/nvim

**Expected behavior**
Auto complete shows up as expected. 

**Screenshots**
Image

**`templ info` output**

```shell
(✓) go [ location=/usr/local/go/bin/go version=go version go1.24.0 linux/amd64 ]
(✓) gopls [ location=/home/phenpessoa/.local/share/nvim/mason/bin/gopls version=golang.org/x/tools/gopls v0.18.1 ]
(✓) templ [ location=/home/phenpessoa/go/bin/templ version=v0.3.833 ]
```

 - OS: linux
 - templ CLI version v0.3.833
- Go version go1.24.0 linux/amd64
- `gopls` version golang.org/x/tools/gopls v0.18.1

a-h:
Closes #1086


joerdav:
Had another crack at reading it this morning.

The changes make sense, and I understand how it solves the problem with partial parses.

I worry that this parsing behaviour is now a bit less intuitive? Can we do something using custom error types or return a ParseResponse enum type?

As you read the return values at a certain place in the code, I think it could be quite hard for anyone else to wrap their heads around.


Aside:

Will this interact with gopls properly? If we fail to generate then the gopls representation of the code will be out of sync I imagine? Just thinking if we need to make sure it's blatantly obvious when editing that this has occured.

joerdav:
I wonder if a more formal `ParserResponse` might make this easier to grok?

```go
var elementOpenTagParser = parse.Func(func(pi *parse.Input) (eo parse.Response[elementOpenTag], err error) {
```

It could contain a field for parsing errors (expected errors). Alongside the response itself, and the `Matched` field.

Then the `error` would be reserved for real errors.


A set of responce constructors might then in turn make the code more readable?

```go
return parser.PartialMatch(element), err
...
return parser.Match(element), err
...
return parser.NoMatch(), err
```

This may add too much complexity though. I think renaming `ok` would be a step in the right direction.

PR #1158, opened 2025-05-21
chore: document expression keys

joerdav:
@delaneyj @a-h Made a first crack on the docs.

Writing it out made me realised how unexpected the behaviour of the special attributes are. 

One option would be to not support these cases in dynamic keys (return an _unsafe key in it's place at runtime). Though. I'm not sure how much better a runtime error is than escaping being bypassed!

### DIFF
--- a/docs/docs/03-syntax-and-usage/03-attributes.md
+++ b/docs/docs/03-syntax-and-usage/03-attributes.md
@@ -125,6 +125,28 @@ templ component() {
 <hr style="padding: 10px" class="itIsTrue" />
 ```
 
+## Attribute key expressions
+
+Use a string expression to dynamically set the key of an attribute.
+
+```templ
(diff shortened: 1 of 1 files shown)


delaneyj:
Sorry which attributes are special?

a-h:
The `on*` (e.g. `onclick`) and `class` attributes are special, in that expression values (`name={ this_bit }`) for those attributes are not expected to implement the `templ.stringable` interface, but are expected to implement `templ.CSSClass` or similar.

The reason for this is that expressions can be formed of strings that are under the control of a user, not the developer of the system. Since they could contain vulne

The original design of templ was to make sure that any string that isn't a compile-time constant goes through escaping, and sanitization, however, @joerdav is pointing out that the dynamic attributes don't have the same sanitization applied.

PR #1164, opened 2025-05-29
fix: add nonce attribute to style tag

### LINKED ISSUE #1163: Nonce should be set on CSS elements
Nonces are currently not applied to CSS elements. I'm setting a nonce on the context in my templ application, and would expect it to be applied to all elements generated via CSS elements, however, it is not. This should be fixed.

poly2it:
Fixes #1163.


a-h:
Thanks, much appreciated.

PR #1166, opened 2025-05-31
Update URLs from cosmtrek/air to air-verse/air

philip-peterson:


a-h (approved):
Thanks, looks good. I noticed that the docs use quite an old version of Air, have you tested

PR #1171, opened 2025-06-01
SafeURL with error

### LINKED ISSUE #1167: form action, anchor href should allow (SafeURL, error)
**Before you begin**

Release one hour ago:

templ version
v0.3.887

**Describe the bug**

form action and anchor href should allow (SafeURL, error), not just SafeURL

**To Reproduce**
A small, self-contained, complete reproduction, uploaded to a GitHub repo, containing the minimum amount of files required to reproduce the behaviour, along with a list of commands that need to be run. Keep it simple.

To reproduce, run `templ generate` on this example:

```templ
package main

import "errors"

templ hello() {
	<form action={ foo() }></form>
	<a href={ foo() }></a>
}

func foo() (templ.SafeURL, error) {
	return templ.URL("/foo"), errors.New("bar")
}
```

**Expected behavior**

No compilation error, but instead the generated code assigns the result directly to one var:

```go
		templ_7745c5c3_Err = templruntime.WriteString(templ_7745c5c3_Buffer, 1, "<form action=\"")
			return templ_7745c5c3_Err
		}
		var templ_7745c5c3_Var2 templ.SafeURL = foo() <--- WRONG
		_, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString(templ.EscapeString(string(templ_7745c5c3_Var2)))
			return templ_7745c5c3_Err
		}
		templ_7745c5c3_Err = templruntime.WriteString(templ_7745c5c3_Buffer, 2, "\"></form><a href=\"")
			return templ_7745c5c3_Err
		}
		var templ_7745c5c3_Var3 templ.SafeURL = foo() <--- WRONG
		_, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString(templ.EscapeString(string(templ_7745c5c3_Var3)))
			return templ_7745c5c3_Err
		}
```

jackielii:
fixes #1167

Thi PR changes form action and anchor href to use SafeURL with error


jackielii:
@a-h @joerdav For templ code

```templ
	<form action={ safeUrl() }>With Error</form>
```
The generated code looks like this:
https://github.com/a-h/templ/blob/49392bf29b84816a4032808826c992b082a77a3d/generator/test-form-action/template_templ.go#L63-L67

Can I propose to remove the `[templ.SafeURL]` restriction? So that the generated code may look like:

```go
 var templ_7745c5c3_Var4 string 
 templ_7745c5c3_Var4, templ_7745c5c3_Err = templ.JoinStringErrs(safeUrl()) 
 	return templ.Error{Err: templ_7745c5c3_Err, FileName: `generator/test-form-action/template.templ`, Line: 7, Col: 25} 
 } 
```

This will allow both `string` and `templ.SafeURL` to work. I.e. it would accept both of these functions:
https://github.com/a-h/templ/blob/49392bf29b84816a4032808826c992b082a77a3d/generator/test-form-action/template.templ#L11-L17

docs/docs/03-syntax-and-usage/03-attributes.md
```diff
@@ -180,48 +180,50 @@ templ usage() {
...
+:::tip
+Non-standard HTML attributes can contain URLs, for example HTMX's `hx-*` attributes).
 
-For use on non-standard HTML elements (e.g. HTMX's `hx-*` attributes), convert the `templ.URL` to a `string` after sanitization.
+To sanitize URLs in that context, use the `templ.URL(urlString)` function.
 
 ```templ
 templ component(contact model.Contact) {
-  <div hx-get={ string(templ.URL(fmt.Sprintf("/contacts/%s/email", contact.ID)))}>
+  <div hx-get={ templ.URL(fmt.Sprintf("/contacts/%s/email", contact.ID)) }>
```
jackielii, line 208: According to the description above, this should just be `fmt.Sprintf("/contacts/%s/email", contact.ID)`, URL sanitising is automatic

a-h: It's not automatic for `hx-get` because it's a non-standard attribute.

templ could have a list of "well known" attribute names that contain URLs, but since data-* URLs can contain them, it wouldn't be much use.

So, I think this documentation is accurate.

PR #1177, opened 2025-06-03
feat: Strip superfluous whitespace from class lists

### LINKED ISSUE #1074: strip superfluous spaces from class lists
Consider the following snippet from a templ template:
```html
<div
	class="
		flex w-full h-full
		justify-center items-center
		hover:bg-blue-50
	"
>
```

I often follow this styling to avoid putting very long lists of classes in a single line:

```html
<div
	class="flex w-full h-full justify-center items-center hover:bg-blue-50"
>
```

The length of real class lists in my code is often several times longer.

templ `v0.3.833` generates the following Go code:

```html
<div class=\"\n\t\t\t\t\t\t\tflex w-full h-full\n\t\t\t\t\t\t\tjustify-center items-center\n\t\t\t\t\t\t\thover:bg-blue-50\n\t\t\t\t\t\t\">
```

I believe these line-breaks and tabs serve no purpose and can be stripped.
- This would reduce the size of the generated HTML at runtime.
- ...and make generated code smaller and easier to read.

```html
<div class=\"flex w-full h-full justify-center items-center hover:bg-blue-50\">
```
https://go.dev/play/p/ejEBGL99OAf

brayden-aldrich:
Closes #1074


parser/v2/elementparser.go
```diff
@@ -167,6 +167,11 @@ var (
 			return attr, false, nil
 		}
 
+		// Strip superfluous whitespace characters from css class lists.
+		if k, singleQuote := attr.Key.(ConstantAttributeKey); singleQuote && strings.ToLower(k.Name) == "class" {
```
joerdav, line 171:
```suggestion
		if k, singleQuote := attr.Key.(ConstantAttributeKey); singleQuote && strings.EqualFold(k.Name, "class") {
```

a-h: I think the name "singleQuote" is wrong too.

```
if k, isConstantAttributeKey := attr.Key.(ConstantAttributeKey); isConstantAttributeKey && strings.EqualFold(k.Name, "class") {
```

brayden-aldrich: Thanks for the feedback, both fixes have been pushed!

a-h:
The issue here is that all templ files need to be formatted in the repo, and string stripping whitespace changes the content of the template, hence it showing up as different.

If the test tested the parser behaviour instead, which is what seems to have changed, maybe it would be easier?

PR #1187, opened 2025-06-10
refactor: replace document open count with package reference count in lazy loader

prathshenoy:
In the lazy loader, track the reference count of each package rather than the number of times each document has been opened. The latter stores redundant information since documents in a package are never opened individually. Rather, any document in a package is always opened in conjunction with all other documents in the same package. Managing reference counts at the package level simplifies the logic and avoids unnecessary redundancy.


joerdav:
Cool, yep that makes sense, thanks!

joerdav (approved):

PR #1191, opened 2025-06-12
Refactor cancellation logic into a separate section

philip-peterson:
* Since the logic is semantically a separate section, moves cancellation into a new function.
* Because variables can be moved closer to their usages, moves a couple variable declarations down.

### DIFF
--- a/cmd/templ/generatecmd/cmd.go
+++ b/cmd/templ/generatecmd/cmd.go
@@ -139,14 +139,9 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
 	errs := make(chan error)
 	// Tracks whether errors occurred during the generation process.
 	var errorCount atomic.Int64
-	// For triggering actions after generation has completed.
-	postGeneration := make(chan *GenerationEvent, 256)
 	// Used to check that the post-generation handler has completed.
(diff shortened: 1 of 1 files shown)


cmd/templ/generatecmd/cmd.go
```diff
@@ -135,23 +135,54 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
...
+			slog.Int64("errorCount", errorCount.Load()),
+		)
+		fileEvents := make(chan fsnotify.Event)
+		go func() {
+			if err := watcher.WalkFiles(ctx, cmd.Args.Path, cmd.WatchPattern, fileEvents); err != nil {
+				cmd.Log.Error("Post dev mode WalkFiles failed", slog.Any("error", err))
+				errs <- FatalError{Err: fmt.Errorf("failed to walk files: %w", err)}
+				return
+			}
+			close(fileEvents) // TODO should this be defer'red?
```
philip-peterson, line 168: Should this close be in a defer, since an error could cause an early return and never close the channel?

a-h:
I've done a major refactor in #1227 to hopefully make the intent of all that code a lot clearer.

PR #1192, opened 2025-06-12
fix: Handle live import declaration changes in lazy loader

prathshenoy:
Synchronize the lazy loader with changes to the import section of a templ file. When the imports in a file change (e.g. new or removed dependencies), we now correctly:
- Detect additions to the import graph and open new packages and their transitive dependencies if necessary
- Detect removals and close packages that are no longer required

This ensures that the loader maintains a consistent set of open templ files in response to live edits to import declarations, and avoids broken import-related bugs.

### DIFF
--- a/cmd/templ/lspcmd/proxy/server.go
+++ b/cmd/templ/lspcmd/proxy/server.go
@@ -245,12 +245,13 @@ func (p *Server) Initialize(ctx context.Context, params *lsp.InitializeParams) (
 	}
 
 	if p.NoPreload {
-		p.templDocLazyLoader = newTemplDocLazyLoader(
-			templDocHooks{
+		p.templDocLazyLoader = newTemplDocLazyLoader(newTemplDocLazyLoaderParams{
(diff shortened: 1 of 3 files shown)


mnoah1 (approved):

prathshenoy:
@joerdav @a-h Could we please get a review of this? We're hoping to get this rolled out by the end of this month. More than happy to discuss any additional details if necessary. Thanks again!

joerdav (changes requested):
I left a very small nit, but besides that it looks good to me. I tested this out on my small bazel project, but I assume you guys have done some testing too?

prathshenoy:
Thanks for the review! We've done thorough testing internally. Were you able to get it working on your Bazel project?

This works well for us, and while there are performance optimizations we could make, our current focus is just getting something functional 🙂.

joerdav:
Yes thanks, appears to be working on my example Bazel project. I get completions once when a new import is added.

joerdav (approved):

PR #1195, opened 2025-06-16
feat: pass stdin to watched commands

### LINKED ISSUE #1194: templ generate with --cmd should pass os.Stdin
**Before you begin**
Please make sure you're using the latest version of the templ CLI (`go install github.com/a-h/templ/cmd/templ@latest`), and have upgraded your project to use the latest version of the templ runtime (`go get -u github.com/a-h/templ@latest`)

Running on head of main: 9b34663a

**Describe the bug**

Run `templ generate --include-version=false -cmd "go run main.go"` when `main.go` contains code that reads terminal. E.g.

```go
import "golang.org/x/term"

if term.IsTerminal(0) {
	width, _, err := term.GetSize(0)
	println("Terminal width:", width)
}
```

This will print 0 because terminal can't be detected

**To Reproduce**
A small, self-contained, complete reproduction, uploaded to a GitHub repo, containing the minimum amount of files required to reproduce the behaviour, along with a list of commands that need to be run. Keep it simple.

See above reproducible example.

**Expected behavior**

`term.IsTerminal(0)` should return true

jackielii:
Fixes #1194

### DIFF
--- a/cmd/templ/generatecmd/run/run_unix.go
+++ b/cmd/templ/generatecmd/run/run_unix.go
@@ -75,6 +75,7 @@ func Run(ctx context.Context, workingDir string, input string) (cmd *exec.Cmd, e
 	cmd.WaitDelay = time.Second * 3
 	cmd.Env = os.Environ()
 	cmd.Dir = workingDir
+	cmd.Stdin = os.Stdin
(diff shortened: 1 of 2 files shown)


a-h:
Thanks!

PR #1197, opened 2025-06-17
refactor: allow more flexible parsing of expression attributes

a-h:
- **chore: bump version [no-ci]**
- **refactor: allow more flexible parsing of expression attributes**


joerdav (approved):
This looks good to me. I agree this change makes sense. It brings us in line with the HTML Standard: 

> The [attribute name](https://html.spec.whatwg.org/#syntax-attribute-name), followed by zero or more [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace), followed by a single U+003D EQUALS SIGN character, followed by zero or more [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace), followed by a single U+0022 QUOTATION MARK character ("), followed by the [attribute value](https://html.spec.whatwg.org/#syntax-attribute-value), which, in addition to the requirements given above for attribute values, must not contain any literal U+0022 QUOTATION MARK characters ("), and finally followed by a second single U+0022 QUOTATION MARK character (").

Issue #1091, opened 2025-03-08
Should all hrefs require templ.SafeURL to be consistent?


drognisep:
**Before you begin**
- [x] Please make sure you're using the latest version of the templ CLI (`go install github.com/a-h/templ/cmd/templ@latest`), and have upgraded your project to use the latest version of the templ runtime (`go get -u github.com/a-h/templ@latest`)
  - I tried using the latest CLI and runtime version, and I'm seeing the same behavior.

**Describe the bug**

First off, I'm not sure if this is a bug. I would expect that a link's href attribute would require a `templ.SafeURL` to be consistent with an anchor tag, but it seems to require a string instead.

I'm using a couple functions to add a URL path prefix to templ templates, and I noticed that static assets in the `head` element referenced by `link` allow (actually require due to how it's generated) a string.

[One function](https://github.com/saylorsolutions/goth-stack/pull/1/files#diff-63036bf427d4f8dd4909e32b2a13809cd96b94b6bae187af7fbb17d3d499523aR20) returns a string (intended for HTMX attributes), [and another](https://github.com/saylorsolutions/goth-stack/pull/1/files#diff-63036bf427d4f8dd4909e32b2a13809cd96b94b6bae187af7fbb17d3d499523aR16) returns a `templ.SafeURL`. I'm using the first variant without issue in [link hrefs](https://github.com/saylorsolutions/goth-stack/pull/1/files#r1985755275).

**To Reproduce**

Add a `link` element to a template (rel icon or stylesheet is what I'm seeing), and pass a plain Go string in curly braces (`{}`). There doesn't seem to be any requirement for a `templ.SafeURL`.

[This is how I've produced this.](https://github.com/saylorsolutions/goth-stack/blob/e8a69ffde045e9fc7df6e258446899990a3135b1/cmd/yourapp/internal/templates/base.templ#L12-L13)

**Expected behavior**

I'd expect consistent behavior between hrefs for different elements, but there could be a good reason for the current behavior that I don't know about.

**Edit:** Trimming down text for clarity and adding specific links.

joerdav:
This makes sense for consistency. I gree that the behaviour would make sense to be the same.

My concern here is backwards compatability, I can imagine that using dynamic strings in a link href is quite niche, I could still see us breaking some code if we introduced a type change. This would potentially have to go in an experiment flag with a warning of some kind if we went ahead.

drognisep:
[...]

PR #1206, opened 2025-06-27
refactor: move expressionAttributeValueURL check to separate function rather than have it inline

grqphical:
based on #1091

I replaced the inline check in `generator.go`
https://github.com/a-h/templ/blob/15199ff1f8fc2cf6a89f5e21c95d642760735645/generator/generator.go#L1347-L1350

With a separate check function `isExpressionAttributeValueURL` that can be extended to check for any combination of an HTML element type and name of an HTML attribute

The function uses a `map[string]string` to store the element/attribute combinations so it is very easy to add more if needed.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -1344,7 +1344,7 @@ func (g *generator) writeExpressionAttribute(indentLevel int, elementName string
 	}
 	attrKey := html.EscapeString(attr.Key.String())
 	// Value.
-	if (elementName == "a" && attrKey == "href") || (elementName == "form" && attrKey == "action") {
+	if isExpressionAttributeValueURL(elementName, attrKey) {
 		if err := g.writeExpressionAttributeValueURL(indentLevel, attr); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


a-h (approved):

a-h:
```go
package main

import (
	"testing"
)

func isExpressionAttributeValueURL(elementName, attrName string) bool {
	var elementAttributeLookup = map[string]string{
		"a":      "href",
		"form":   "action",
		"link":   "href",
		"object": "data",
	}
	expectedAttribute, exists := elementAttributeLookup[elementName]
	if !exists {
		return false
	}
	return attrName == expectedAttribute
}

func isExpressionAttributeValueURL2(elementName, attrName string) bool {
	switch elementName {
	case "a", "link":
		return attrName == "href"
	case "form":
		return attrName == "action"
	case "object":
		return attrName == "data"
	default:
		return false
	}
}

var elementAttributeLookup3 = map[string]string{
	"a":      "href",
	"form":   "action",
	"link":   "href",
	"object": "data",
}

func isExpressionAttributeValueURL3(elementName, attrName string) bool {
	expectedAttribute, exists := elementAttributeLookup3[elementName]
	if !exists {
		return false
	}
	return attrName == expectedAttribute
}

func BenchmarkIsExpressionAttributeValueURL(b *testing.B) {
	b.ReportAllocs()
	for b.Loop() {
		_ = isExpressionAttributeValueURL("a", "href")
	}
}

func BenchmarkIsExpressionAttributeValueURL2(b *testing.B) {
	b.ReportAllocs()
	for b.Loop() {
		_ = isExpressionAttributeValueURL2("a", "href")
	}
}

func BenchmarkIsExpressionAttributeValueURL3(b *testing.B) {
	b.ReportAllocs()
	for b.Loop() {
		_ = isExpressionAttributeValueURL3("a", "href")
	}
}
```

```
0 0 /Users/adrian/github.com/a-h/templbm % go test -bench .
goos: darwin
goarch: arm64
pkg: templbm
cpu: Apple M1 Max
BenchmarkIsExpressionAttributeValueURL-10       22368434                53.52 ns/op            0 B/op          0 allocs/op
BenchmarkIsExpressionAttributeValueURL2-10      580624324                2.068 ns/op           0 B/op          0 allocs/op
BenchmarkIsExpressionAttributeValueURL3-10      148611994                8.183 ns/op           0 B/op          0 allocs/op
PASS
ok      templbm 3.796s
```

So, the 2nd one is much faster.

PR #1216, opened 2025-07-09
feat: add fragment rendering support

a-h:
This PR adds fragment rendering support.

It enables an optimisation for HTMX by returning a smaller HTML payload when an entire document is not required. When carrying out a `hx-get` or similar, if an entire page is returned the  `hx-select` attribute (https://htmx.org/attributes/hx-select/) needs to be used to select the subsection of content to replace. This enables the server to return only the requested content.

The fragment concept is discussed further in https://htmx.org/essays/template-fragments/

I've included a docs update, a HTMX example, tests for the HTTP handler, and a set of generator tests which can further explain the feature.


joerdav:
This looks good! I think this will be very useful!

2 notes:

- Should we document and define the behavour if 2 fragments exist with the same name?
- I wonder if it would be useful to have a `func RenderFragment(context.Context, templ.Component, string, io.Writer) error` to cover the 90% of cases for this feature?

joerdav:
Yep, that behaviour was intuitively what I expected when I read the docs, so makes sense!

If we go with the helper function, does it make sense to make the context functions private, shrinking our API surface? Only expose the RenderFragment and WithFragment functions for now?

Maybe to match the other function signatures the helper should actually be:

`func RenderFragment(context.Context, io.Writer,, templ.Component string...) error`

PR #1225, opened 2025-07-23
feat: add Ranges to GoComment parser nodes

dgrundel:
Hi there, I'm working on a Templ Linting tool for my project. In order to (a) display useful results to users and (b) enable linter config via comments (similar to eslint's `eslint-disable`), the tool needs to be able to understand the original source location of parsed nodes.

I came across a discussion thread on this topic and @joerdav seemed to agree that Templ's own parser should be able to support this: https://github.com/a-h/templ/discussions/586

So, here I am with what I hope to be the first of a handful of pull requests to add `Range`s to various parser nodes. I kept this one small to ensure that I'm following the right conventions, etc. Happy to receive feedback on the PR. I would like to make future PRs a bit longer so I'm not bombarding you with them, but if you prefer many smaller changes that'd fine with me.

Thanks for taking a look!

### DIFF
--- a/parser/v2/gocommentparser_test.go
+++ b/parser/v2/gocommentparser_test.go
@@ -21,6 +21,7 @@ func TestGoCommentParser(t *testing.T) {
 				Contents:  " single line comment",
 				Multiline: false,
 				Range: Range{
+					From: Position{Index: 0, Line: 0, Col: 0},
 					To:   Position{Index: 22, Line: 0, Col: 22},
 				},
 			},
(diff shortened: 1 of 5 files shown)


parser/v2/gocommentparser_test.go
```diff
@@ -20,6 +20,9 @@ func TestGoCommentParser(t *testing.T) {
 			expected: &GoComment{
 				Contents:  " single line comment",
 				Multiline: false,
+				Range: Range{
+					To:   Position{Index: 22, Line: 0, Col: 22},
```
a-h, line 24: I know that the From value is a zero value, but I think it would be clearer if `From: Position{Index: 0, Line: 0, Col: 0},` was visible.

a-h (changes requested):
Looks good, just a minor nit if you don't mind, then happy to merge.

A number of small PRs would be preferable to a big one.

parser/v2/gocommentparser_test.go
dgrundel: Agreed! Thank you for updating and merging. I will be sure to do this for future PRs.

PR #1229, opened 2025-07-27
feat: add Range to HTMLComment nodes

dgrundel:
Hi again, this adds a `Range` to the parser's `HTMLComment` nodes. 

**Note on testing:** 
I am planning to stick to reusing the existing unit tests and to just update to include ranges. I'm finding that the current set of tests have a good range of inputs. If I spot any missing edge cases, I'll add more tests. Obviously, I'm happy to add any additional tests that you feel are needed as well. Just let me know if you see any gaps.

**Use case notes:**
As [noted previously](https://github.com/a-h/templ/pull/1225), my particular use case for these ranges is for linting Templ code. With each of these PRs I'll add a few words on why a range is useful for this particular element (i.e. why would you want to lint this?)

In the case of HTML comments, you may want to create a rule to forbid them entirely, or at least to ensure that they do not contain commented code. In addition to bloating the payload you send to your visitors, HTML comments may also leak unnecessary information. In most cases -- at least in my organization -- Go comments are preferred.

### DIFF
--- a/parser/v2/htmlcommentparser.go
+++ b/parser/v2/htmlcommentparser.go
@@ -35,5 +35,6 @@ func (p htmlCommentParser) Parse(pi *parse.Input) (n Node, ok bool, err error) {
 		return
 	}
 
+	c.Range = NewRange(start, pi.Position())
 	return c, true, nil
 }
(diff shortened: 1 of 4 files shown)


joerdav:
Great, this looks good thanks!

Issue #481, opened 2024-02-01
question: Does/Should javascript blocks be formatted with `templ fmt` command?


jmarais:
This is just a question as per the title. I am not sure if it should work.

example file with weird indentations:

```bash
$ templ fmt < ./siteexp.templ
package templates

script graph(data string) {
        const chart = LightweightCharts.createChart(document.body, { width: 400, height: 300 });
                const lineSeries = chart.addLineSeries();
        lineSeries.setData(data);
    if (true) {
    console.log("yes");
    } else {
    console.log("no");
    }
}

templ body() {
        <script>
                const chart = LightweightCharts.createChart(document.body, { width: 400, height: 300 });
                const lineSeries = chart.addLineSeries();
                lineSeries.setData([
                                { time: '2019-04-11', value: 80.01 },
                                { time: '2019-04-12', value: 96.63 },
                                        { time: '2019-04-13', value: 76.64 },
                                { time: '2019-04-14', value: 81.89 },
                                { time: '2019-04-15', value: 74.43 },
                                        { time: '2019-04-16', value: 80.01 },
                                { time: '2019-04-17', value: 96.63 },
                                { time: '2019-04-18', value: 76.64 },
                                { time: '2019-04-19', value: 81.89 },
                                        { time: '2019-04-20', value: 74.43 },
                ]);
        </script>
        <body>
                <div>
                        hello
                </div>
        </body>
}

```

The html 'body' section at the end does format, however nothing in either a 'script' block or tags gets formatted.

joerdav:
Hi @jmarais to answer your question of if templ "does" any formatting, it doesn't currently no.

Whether it should is a harder question, there are a lot of differing opinions on how formatting should be done in terms of JS, so my suggestion would be to encourage separate tools that allow users to configure preferences, rather than having these baked in to templ.

jmarais:
Thanks

a-h:
Interesting idea. I think that templ could detect the presence of a .prettierrc and, if present, and the prettier executable is on the path, could call out and format the JS.

If a .prettierrc was found, but there's no prettier executable, then it could be a warning.

JonnyLoughlin:
If you are using Neovim, you can run your formatters with https://github.com/stevearc/conform.nvim, which has injected language support through tree sitter. I use it to run biome on the JavaScript portions of my code.

jmarais:
@JonnyLoughlin, Thanks. I will give conform a shot.

JonnyLoughlin:
[...]

PR #1230, opened 2025-07-27
feat: format js and css with prettier if prettier is on the $PATH

a-h:
Closes https://github.com/a-h/templ/issues/481


therealparmesh:
Looks good!

Hopefully this replaces my little hack: https://github.com/therealparmesh/prettier-plugin-templ-script

Will this be compatible with tailwind’s prettier plugin for spring classes?

a-h:
@therealparmesh - not at the moment, because it's only formatting the contents of `<script>` and `<style>` elements, but the approach could be extended to apply prettier rules to constant attribute values to sort class names.

internal/prettier/prettier.go
```diff
@@ -0,0 +1,45 @@
...
+	"os/exec"
+	"strings"
+	"sync"
+)
+
+var resolveFormatter = sync.OnceFunc(func() {
+	candidates := [][]string{
+		{"prettierd"},
+		{"prettier"},
+		{"npx", "prettier"},
```
therealparmesh, line 15: I wonder if this can be expanded some, perhaps via configuration. Let's say the dev has a `frontend` folder in their repository:

```tree
my-app/
├─ templates/
│  ├─ root.templ
├─ go.mod
├─ main.go
├─ frontend/
│  ├─ node_modules/
│  ├─ package.json
│  ├─ prettier.config.js
```

If we give them a way to pass in the dir, prettier can be executed in that context, with all plugins that they might have.

a-h: Sounds reasonable, something like`templ fmt --prettier-cmd="npx prettier" --prettier-config=./frontend`?

therealparmesh: `--use-tabs` is an interesting tidbit, because it really should be enabled (like you have it here), but maybe it's just a bit of documentation.

internal/format/templ.go
```diff
@@ -0,0 +1,164 @@
...
+	case *parser.RawElement:
+		nodeToDepth[e] = depth
+	case parser.CompositeNode:
+		for _, child := range e.ChildNodes() {
+			calculateNodeDepth(child, nodeToDepth, depth+1)
+		}
+	}
+}
+
+func prettifyElement(name string, typeAttrValue string, content string, depth int, prettierCommand string) (after string, err error) {
```
joerdav, line 103: The indentation side of this function feels very much like behavior closely linked to prettier. Should it live in the prettier package?

PR #1231, opened 2025-07-28
feat: add Range to raw elements (RawElement, ScriptElement)

dgrundel:
Hi, back again! This time adding Range to the elements included in the `rawElements` parser: `RawElement` and `ScriptElement`.

**Testing:** Here again I just updated the existing tests. Testing is a little slim for these elements, let me know if you'd like to see additional cases added.

**Use case notes:** In my organization we generally ban script and style tags directly in markup, preferring CSS and JS served from CDNs. (Or, better yet, no additional CSS at all except for those provided by Tailwind.) In cases where they are allowed, you want to be able to extract the contents of these tags and run them through separate, language-specific tools (e.g. stylelint, eslint). Lastly, there's a common pattern where server-side state is emitted into the page by inserting a script tag (with `type` like `application/json`) containing only a JSON payload. For these we'd want to ensure a consistent `type` attribute, perhaps require an `id`, and ensure that the contents of the tag is only a single go expression.

### DIFF
--- a/parser/v2/raw.go
+++ b/parser/v2/raw.go
@@ -62,5 +62,6 @@ func (p rawElementParser) Parse(pi *parse.Input) (n Node, ok bool, err error) {
 	// Cut the end element.
 	_, _, _ = end.Parse(pi)
 
+	e.Range = NewRange(pi.PositionAt(start), pi.Position())
 	return e, true, nil
 }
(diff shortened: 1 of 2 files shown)


a-h:
Looks good!

re: tests for this section, the parser tests check that the elements can be parsed into the object model well, but there's also formatting tests and generation tests that cover the same area from different angles.

You might be interested in https://github.com/a-h/templ/pull/1230 which uses prettier to format `<script>` and `<style>` elements in templ code.

In that PR, I've also refactored the `fmt` command and behaviour so that it is easier to read and re-use.

PR #1235, opened 2025-08-05
feat: don't insert proxy reload script on datastar initiated requests

a-h:


delaneyj:
I have no idea, I never use the templ proxy as I don't have the need.

I use a Taskfile task
```yaml
  templ:
    env:
      TEMPL_EXPERIMENT: rawgo
    # deps:
    #   - library
    generates:
      - "**/*_templ.go"
    sources:
      - "**/*.templ"
    cmds:
      - templ generate .
 ```
a shared page setup...
```html
<body>
	if site.Env.IsDev {
		<div id="hotreload" data-on-load="@get('/hotreload', {retryMaxCount: 1000,retryInterval:20, retryMaxWaitMs:200})"></div>
	}
	{ children... }
	if site.Env.IsDev {
		<datastar-inspector></datastar-inspector>
	}
</body>
```
and a Go route...
```go
        var hotReloadOnlyOnce sync.Once
	router.Get("/hotreload", func(w http.ResponseWriter, r *http.Request) {
		sse := datastar.NewSSE(w, r)
		hotReloadOnlyOnce.Do(func() {
			sse.ExecuteScript("window.location.reload()")
		})
		<-r.Context().Done()
	})
```
That way I get auto loading of anything from css, images, go, etc so my frontend and backend are always in sync.  If there is something specific you need please let me know but I don't think Datastar needs the proxy at all given the builtin SSE support with auto reconnect.

delaneyj:
> Thanks @delaneyj - hope all is well with you and family by the way!

You as well!  Would be nice to catch up when you have time!

PR #1242, opened 2025-08-10
Update 05-statements.md

assaidy:


assaidy:
The `switch`/`for` examples were already covered earlier in the document, so I removed the redundancy. Instead, I added code focused on resolving the `if` issue in this section.

Issue #1243, opened 2025-08-14
Error parse template error when script has regex in templ "template closing brace not found"


minhluan2292:
Templ parse err when script has **regex**:

`const clientIdMatch = evt.detail.message.match(/data-client-id="([^"]+)"/);
`
Error: **parsing error: template closing brace not found**

but write regex twice will no error: 
```
const regex = /data-client-id="([^"]+)"/;
const clientIdMatch = evt.detail.message.match(/data-client-id="([^"]+)"/);
```
Try this also error as above: 
```
const regex = /data-client-id="([^"]+)"/;
const clientIdMatch = evt.detail.message.match(regex);
```
**Environment**: 
- Go: go version go1.23.3
- templ: v0.3.920

PR #1245, opened 2025-08-15
fix: skip parsing of JavaScript regexp literals: `/test/`

a-h:
Fixes https://github.com/a-h/templ/issues/1243

But introduces an issue in code like this:

```
const result = call(1000 / 10, {{ data }}, 1000 / 10);
```

In the current PR, `/ 100, {{ data }}, 1000 /` would be collected as a regexp literal, and therefore, the `{{ data }}` go value wouldn't be interpolated.


joerdav (approved):
Looks good, it reduces the cases that would error! Definitely would be ideal to squash the double curly pair bug but I think can be a separate pr.

PR #1251, opened 2025-08-18
feat: support Go comments within HTML tag attributes

### LINKED ISSUE #866: proposal: Go comments within HTML tags
In addition to #79, I'd like to propose `//` comments within HTML tags, such as:

```html
<form
  class="flex"
  // This comment here explains why the following attribute is necessary,
  // namely to prevent submit when hitting the enter key.
  // But templ won't allow this comment to be here and throws syntax errors.
  action="javascript:void(0)"
>
```
image

## Current Workaround

Currently, the comments must be put above the HTML tag, which is not always convenient.

```html
// Use action="javascript:void(0)" to prevent submit on enter
<form
  class="flex"
  action="javascript:void(0)"
>
```

Since `//` is not a valid token inside an HTML tag IMHO it should be possible to allow Go comments within tags that aren't included in the _compiled_ HTML.

jackielii:
## Summary
Implements support for single-line (`//`) and multi-line (`/* */`) Go comments within HTML element attributes, addressing issue #866.

## Problem
Previously, templ did not allow Go comments within HTML tag attributes, requiring developers to place comments above the element which wasn't always convenient for documenting specific attributes.

## Solution
This PR adds support for Go-style comments directly within HTML tag attributes. The comments are parsed and stored in the AST but excluded from the generated output.

### Example
```templ
<form
  class="flex"
  // This comment explains why the following attribute is necessary
  action="javascript:void(0)"
  /* Multi-line comment
     for more detailed explanations */
  method="post"
>
```

## Changes
- Added `AttributeComment` type to represent comments within attributes
- Updated parser to recognize and handle comments between attributes  
- Updated generator to skip comments during code generation
- Added comprehensive test coverage in `generator/test-attribute-comments/`

## Testing
- Added test case following the existing test pattern in the generator folder
- Test verifies that comments are properly excluded from generated output
- All existing tests continue to pass

Fixes #866


joerdav (approved):

jackielii:
@a-h anything else you need?

a-h:
@joerdav - we discussed this one last Friday, right?

IIRC, we said we'd merge it.

I was a bit reluctant because it's hard to walk something back once you've added it, and it changes the syntax highlighting and some other stuff.

jackielii:
I've been using this patch over a year now. I'm exclusive Neovim user, the treesitter already supports this. I think it's a natural extension to existing syntax: support commenting inside the HTML tag is less unexpected than not supporting it.

a-h:
Agreed. I'm merging this today, just fixing the conflicts.

PR #1271, opened 2025-09-17
feat(proxy): flush streamed html

ZackarySantana:
### Description

When using streaming (e.g. the [suspense example](https://github.com/a-h/templ/tree/main/examples/suspense)) and the proxy option from `templ generate`, streamed in pages will not properly update. This is because they are loaded into memory completely then proxied.

This pr adds some flushing, by sending over tokens as it reads them.

### Screenshots

Before:
https://github.com/user-attachments/assets/9f15dc67-82ae-4fa6-8730-2688ee8fa07a

After:
https://github.com/user-attachments/assets/de6f51f3-04da-454d-be41-c636c64b53ca

### DIFF
--- a/cmd/templ/generatecmd/proxy/proxy_test.go
+++ b/cmd/templ/generatecmd/proxy/proxy_test.go
@@ -477,12 +477,12 @@ func TestProxy(t *testing.T) {
 		}
 
 		streamingGap := 100 * time.Millisecond
-		var writerErr1, writerErr2 error
+		var writerErr1, writerErr2, reqWriterErr error
 		go func() {
-			defer reqWriter.Close()
 			_, writerErr1 = io.WriteString(reqWriter, `<html><head></head><body>`)
 			time.Sleep(streamingGap) // simulate streaming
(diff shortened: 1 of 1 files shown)


cmd/templ/generatecmd/proxy/proxy.go
```diff
@@ -114,7 +166,43 @@ func (h *Handler) modifyResponse(r *http.Response) error {
 	case "":
 		log.Debug("No content encoding header found")
 	default:
-		h.log.Warn(unsupportedContentEncoding, slog.String("encoding", r.Header.Get("Content-Encoding")))
+		log.Warn(unsupportedContentEncoding, slog.String("encoding", r.Header.Get("Content-Encoding")))
```
ZackarySantana, line 169: This log line using the unwrapped logger seemed unintentional to me, so I changed it to use the wrapped one

ZackarySantana:
I rebased on main to fix a failing test

a-h:
Sorry about the delay, got a bit of a backlog. I will get to this soon, and it sounds like a useful update!

a-h (approved):
Cracking. Thanks.

PR #1272, opened 2025-09-18
ci: typo generate

ZackarySantana:
Runs generate after #1270 broke it

image


a-h:
Ah, thanks - I already did it!

PR #1297, opened 2025-11-09
Datastar & htmx docs improvements

bencroker:
This PR fixes the docs page for Datastar, which has new installation instructions, an updated syntax, and a new import path.

Also renames `HTMX` to `htmx`, which is the official spelling.

### DIFF
--- a/docs/docs/05-server-side-rendering/04-datastar.md
+++ b/docs/docs/05-server-side-rendering/04-datastar.md
@@ -76,7 +76,7 @@ templ templCounterExampleInitialContents(signals TemplCounterSignals) {
 ```
 
 :::tip
-Note that Datastar sends all[^1] signals to the server (as JSON) on each request. This means far less bookkeeping and more predictable state management.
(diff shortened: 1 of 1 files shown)


docs/docs/05-server-side-rendering/04-datastar.md
```diff
@@ -88,15 +76,15 @@ templ templCounterExampleInitialContents(signals TemplCounterSignals) {
 ```
 
 :::tip
-Note that Datastar doesn't promote the use of forms because they are ill-suited to nested reactive content. Instead, it sends all[^1] reactive state (as JSON) to the server on each request. This means far less bookkeeping and more predictable state management.
+Note that Datastar sends all[^1] signals to the server (as JSON) on each request. This means far less bookkeeping and more predictable state management.
```
nickchomey: Can I suggest 
```suggestion
Note that Datastar sends all[^1] signals to the server (as JSON) on each request. This means far less bookkeeping and more predictable state management than when using html forms. 
```

bencroker: Thanks!

joeaviva (approved):

PR #1302, opened 2025-11-14
feat: add Range to DocType nodes

AdamVig:
Adds a `Range` to the parser's `DocType` nodes.

Continues @dgrundel's initiative of adding `Range` to all AST nodes for the benefit of external linting tools, which started in #1225 and was originally discussed in https://github.com/a-h/templ/discussions/586.

See also #1301.


a-h (approved):

a-h:
Thanks!

PR #1310, opened 2025-12-11
Configurable Formatting

Treppord:
Format config

### DIFF
--- a/cmd/templ/fmtcmd/main.go
+++ b/cmd/templ/fmtcmd/main.go
@@ -16,20 +16,28 @@ import (
 )
 
 type Arguments struct {
-	FailIfChanged    bool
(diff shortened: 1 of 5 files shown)


joerdav:
Hi there, thankyou for taking your time to contribute to templ.

The goal of the fmt command is to be opinionated, like the go fmt command! So any new rules we add, we wouldn't add behind a configuration flag.

I suggest that for any rules that you think would be valuable for everyone, raise an issue on our tracker so we can discuss whether it is a good fit for templ.

Your other option would be to write a custom formatter. The parser package contains everything you need to parse a templ file, and the parsed types provide `Write` functions to re-write the formatted node to a templ file (you would override this in any cases that you would like to replace the formatting rules.

PR #1321, opened 2026-01-15
fix: proxy escaping characters

### LINKED ISSUE #1316: Inline scripts break in v0.3.977 (&& → &amp;&amp;)
Hey! 

After updating to `v0.3.977`, my inline `<script>` tags get escaped - `&&` becomes `&amp;&amp;`, quotes become `&#39;`, etc.

Rolled back to v0.3.960 and it works fine again.

MilkeeyCat:
The bug was introduced in #1271. `Token`'s `String` method(https://cs.opensource.google/go/x/net/+/master:html/token.go;l=105) escapes tag body text, which shouldn't happen. `String` method is replaced by `Raw` which makes sense imo because the html is already produced by templ, there's no need to format/process it again.

fixes #1316

### DIFF
--- a/cmd/templ/generatecmd/proxy/proxy.go
+++ b/cmd/templ/generatecmd/proxy/proxy.go
@@ -111,8 +111,7 @@ func streamInsertAfterBodyOpen(nonce string, r io.Reader, w io.Writer) error {
 				}
 			}
 		default:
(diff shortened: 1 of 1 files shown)


a-h:
Nice find. Thanks!

PR #1345, opened 2026-02-15
feat: add TLS support to live reload proxy

### LINKED ISSUE #1329: Add HTTPS support for the live-reload proxy
Would it be possible to have the live-reload proxy support HTTPS? For example, testing login flows with login-with-google will fail since google will not allow a redirect url to a non-https url. 

It's a small inconvenience to get redirected back to the non-proxy url, then change the protocol and port in the address bar and add/update session cookie, however doing this multiple times / day / week can add up and also causes frustration when forgetting to do it and wondering why it isn't reloading.

a-h:
Closes #1329


BillBuilt:
I tried making a brand new cert



`mkcert localhost 127.0.0.1 ::1`

But encountered the same error.
image






:8080 is secure 
image

PR #1361, opened 2026-03-23
Adding coverage reporting for Templ files (part 1/3)

whalenrp:
Hey all, 
I was playing around with the SourceMap infrastructure to see if it could support some reverse-engineering of a templ coverage report by taking the Go coverage and mapping it back to templ lines. I quickly realized it wouldn't be as simple as I thought:
1. Sourcemaps seem only implemented to the extend needed to support LSP operations
2. A single templ statement like a spread operator maps to many lines of Go, making it harder to map go coverage back to templ. 
3. Even if these worked, we'd still need custom tooling for generating reports anyway

Instead of trying to make this all work, we could track what lines are executed at runtime in the generated code itself, marking templ lines as they're executed in Go. This involves modifying the generated code to instrument it with some tracking statements and flushing those out when the test finishes execution.

The full workflow for users is:

1. Add `RunWithCoverage` to `TestMain` (one-time setup per test package)
2. Run `templ generate --coverage` before tests //opt-in to capturing tracking information
3. Set `TEMPLCOVERDIR` when running tests // opt-in to writing out the profile info
4. Run `templ coverage report` to see results

I've been playing with this and have a [prototype](https://github.com/a-h/templ/compare/main...whalenrp:templ:coverage-v1-3?expand=1) together. The HTML reporting is not included in that stack, but I wanted to show you how it looks end-to-end 
Screenshot 2026-03-22 at 9 42 00 PM

This diff specifically includes just the code needed to track, collect, and flush out the coverage data if you agree with the overall idea. I'm happy to make any changes in the approach if you all have suggestions of course.

### DIFF
--- a/runtime/coverage.go
+++ b/runtime/coverage.go
@@ -1,0 +1,213 @@
+package runtime
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"sync"
+	"syscall"
+	"time"
+)
+
+// Position represents a source location in a template file
+type Position struct {
+	Line uint32 `json:"line"`
+	Col  uint32 `json:"col"`
+}
+
+// CoveragePoint represents a single coverage measurement
+type CoveragePoint struct {
+	Line uint32 `json:"line"`
+	Col  uint32 `json:"col"`
+	Hits uint32 `json:"hits"`
+	Type string `json:"type"`
+}
+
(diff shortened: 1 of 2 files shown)


whalenrp:
I'm going to close this for now. I think it may be easier to build this sort of tooling on top of the existing sourcemaps now that the AST work is almost done. I'll chip in on that side since there's already prior art there and it could be useful for a number of things (coverage metrics, stack trace processing, linting, etc.)

PR #1370, opened 2026-03-28
fix: use dedicated shadow host in Suspense example to ensure header is rendered

scrhartley:
Fixes #1357.

When the declarative shadow DOM creates a shadow root it makes the element it's in the shadow host. Any existing content is replaced, so add a dedicated shadow host element to avoid accidental missing content (the `h1` in this example).

The alternative to this fix would be to remove the `h1` element entirely and have the `body` element as the intended shadow host.

### DIFF
--- a/docs/docs/05-server-side-rendering/05-streaming.md
+++ b/docs/docs/05-server-side-rendering/05-streaming.md
@@ -119,20 +119,22 @@ templ Page(data chan SlotContents) {
 		</head>
 		<body>
 			<h1>Page</h1>
-			@templ.Flush() {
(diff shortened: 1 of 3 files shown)


a-h (approved):
Thanks!

scrhartley:
Does something need to happen so that the website reflects the change?
https://templ.guide/server-side-rendering/streaming/

a-h:
Ah, yes. The docs are released at the same time as a tag is pushed, so that the docs reflect the latest software, but I can manually run the docs-deploy workflow, so I'll do that.

Thanks!

PR #1375, opened 2026-04-15
fix: allow JSFuncCall on arbitrary HTML attributes

### LINKED ISSUE #1320: JSFuncCall not working correctly on arbitary html attributes
**Describe the bug**
JSFuncCall not working on every attribute. The same expressions fails depending on HTML attrubute name

**To Reproduce**
This does not compile
```templ
templ Test(test string) {
	<div x-data={ templ.JSFuncCall("alert", test) }>test</div>
}

```
output:
```
templ.ComponentScript does not satisfy templ.stringable (templ.ComponentScript missing in ~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr | ~float32 | ~float64 | ~complex64 | ~complex128 | ~string | ~bool)

failed to build, error: exit status 1
```

This does compile:

```templ
templ Test(test string) {
	<div onclick={ templ.JSFuncCall("alert", test) }>test</div>
}

```
Produces the following html (for example)
```html
<div onclick="alert(&quot;test&quot;)">test</div>
```

**Expected behavior**
Templ should compile this to the correct function call.

**Screenshots**
If applicable, add screenshots or screen captures to help explain your problem.

**Logs**
If the issue is related to IDE support, run through the LSP troubleshooting section at https://templ.guide/developer-tools/ide-support/#troubleshooting-1 and include logs from templ

**`templ info` output**
(✓) os [ goos=linux goarch=arm64 ]
(✓) go [ location=/usr/local/go/bin/go version=go version go1.25.5 linux/arm64 ]
(✗) gopls [ location= version= message=failed to find gopls: cannot find gopls on the path ("/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"), in $HOME/go/bin or $HOME/.local/bin/gopls. You can install gopls with `go install golang.org/x/tools/gopls@latest` ]
(✓) templ [ location=/go/bin/templ version=v0.3.977 ]
(!) prettier [ location= version= message=failed to find prettier: exec: "prettier": executable file not found in $PATH ]

 - OS: Linux (Docker Container)
 - templ version v0.3.977
- Go version go1.25.5 linux/arm64
- `gopls` version not used

**Additional context**
In my opinion this is very relevant. In alpine.js it is very elegant to write a function which returns the alpine data like this
```js
function someData(serverdata) {
   if(serverdata){
       return {...serverdata}
   } else {
       return {test:"test"}
   }

}
```

I also tested this with arbitary html attributes which also do not work, therefore other javascript libraries with other html-attributes would not work.

a-h:
Fixes #1320

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -663,6 +663,24 @@ func JoinStringErrs[T stringable](s T, errs ...error) (string, error) {
 	return fmt.Sprint(s), errors.Join(errs...)
 }
 
+// ResolveAttributeValue converts an attribute expression value to an
+// HTML-escaped string suitable for use in an attribute value. It handles
+// ComponentScript values by returning their Call field (already escaped),
+// and falls back to HTML-escaping fmt.Sprint for other types.
+func ResolveAttributeValue(v any, errs ...error) (string, error) {
+	if err := errors.Join(errs...); err != nil {
+		return "", err
+	}
+	switch v := v.(type) {
+	case ComponentScript:
+		return v.Call, nil
+	case string:
+		return EscapeString(v), nil
+	default:
+		return EscapeString(fmt.Sprint(v)), nil
+	}
+}
+
 // Error returned during template rendering.
 type Error struct {
 	Err error
--- a/benchmarks/templ/template_templ.go
+++ b/benchmarks/templ/template_templ.go
@@ -46,11 +46,11 @@ func Render(p Person) templ.Component {
 			return templ_7745c5c3_Err
 		}
 		var templ_7745c5c3_Var3 string
-		templ_7745c5c3_Var3, templ_7745c5c3_Err = templ.JoinStringErrs(`something with "quotes" and a <tag>`)
+		templ_7745c5c3_Var3, templ_7745c5c3_Err = templ.ResolveAttributeValue(`something with "quotes" and a <tag>`)
 		if templ_7745c5c3_Err != nil {
 			return templ.Error{Err: templ_7745c5c3_Err, FileName: `benchmarks/templ/template.templ`, Line: 6, Col: 104}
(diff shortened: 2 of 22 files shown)


runtime.go
```diff
@@ -663,6 +663,24 @@ func JoinStringErrs[T stringable](s T, errs ...error) (string, error) {
 	return fmt.Sprint(s), errors.Join(errs...)
 }
 
+// ResolveAttributeValue converts an attribute expression value to an
+// HTML-escaped string suitable for use in an attribute value. It handles
+// ComponentScript values by returning their Call field (already escaped),
+// and falls back to HTML-escaping fmt.Sprint for other types.
+func ResolveAttributeValue(v any, errs ...error) (string, error) {
```
shYkiSto, line 670: @a-h One side effect of this change is that it relaxes the accepted Go types for default expression attributes:
* Before this patch, default expression attributes were generated through `templ.JoinStringErrs()`, so the value had to satisfy the string-like constraint: https://github.com/a-h/templ/blob/8fecf2d46b65db09ef5447594cdfea52c3ecaed0/runtime.go#L657-L664
* Which now generates `templ.ResolveAttributeValue()`, which accepts `any`. That relaxes the input type, so other non-string values now compile too, e.g. maps, structs, etc.

Should `ResolveAttributeValue` be narrowed to accept the previous `stringable` set plus `ComponentScript`, or was this broader type change intentional?

a-h: Oh yeah, that was unintentional. Good spot.

PR #1381, opened 2026-04-29
feat: add prettier formatting to LSP

jhzn:
I want the LSP to run prettier, like what `templ fmt` does.

This solution:
https://templ.guide/developer-tools/ide-support#formatting
is not good enough to standardize around for a team using different editors. 

I threw this PR together as a working example but I'm very open to suggestions.


a-h:
Thanks, I'll take a look!

PR #1395, opened 2026-05-14
fix: failure to parse large templ files

akfaew:
Fix templ parser failure on large `.templ` files

This fixes a parser failure where large `.templ` files could report a misleading error near the top of the file, for example:

`templ: malformed templ expression, expected templ functionName() {: line 3, col 0`

## Root cause

`parseGoFuncDecl` passed the entire remaining `.templ` source into the Go expression parser when extracting a templ function declaration. For large files, that meant the parser also saw all following templ/HTML content after the current declaration, and Go parser recovery could fail before templ got to the real content.

## Fix

Scan only the templ function declaration prefix, up to the opening brace of the function body, then parse that as a normal Go function declaration.

This keeps the Go parser focused on:

`func Name(...) {}`

instead of the entire rest of the `.templ` file.

## Tests

Added a regression test with many generated templ components to cover large files with substantial trailing content.

Verified with:

`go test ./parser/v2/...`

### DIFF
--- a/parser/v2/goparser.go
+++ b/parser/v2/goparser.go
@@ -15,7 +15,7 @@ func parseGoFuncDecl(prefix string, pi *parse.Input) (name string, expression Ex
 	from := pi.Index()
 	src, _ := pi.Peek(-1)
 	src = strings.TrimPrefix(src, prefix)
-	decl, err := funcDeclSource(src)
+	decl, err := extractFuncDeclSignature(src)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


a-h:
Good idea to focus the parser on a smaller set of code. I'll look at this in detail.

I want to do a bit of testing on the impact of setting up the fileset etc. and explore the potential for parsing errors with complex function definitions (funcs that take anonymous functions, generics etc.).

But looks like a good starting point.

a-h (approved):

PR #1398, opened 2026-05-20
feat: track element tag ranges

AdamVig:
Adds `OpenTagRange`, `CloseTagRange`, and `SelfClosing` to element types. Also adds `NameRange` to `RawElement`.

Continues the initiative of adding `Range` to all AST nodes for the benefit of external linting tools, which started in #1225 and was originally discussed in https://github.com/a-h/templ/discussions/586.

See also #1301, #1302, #1335, #1336, #1337, #1338, #1340, #1341, #1347, #1348, #1350, and #1383.

### DIFF
--- a/parser/v2/elementparser.go
+++ b/parser/v2/elementparser.go
@@ -12,11 +12,12 @@ import (
 
 // Element open tag.
 type elementOpenTag struct {
-	Name        string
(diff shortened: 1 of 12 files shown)


a-h (approved):
Thanks. Nice change on Void -> SelfClosing too.

PR #1409, opened 2026-06-12
feat: add -no-format flag to skip formatting on generated output

whalenrp:
When set, the generator writes raw output directly without the format step. Source map positions are then consistent with the generated file content, which enables programmatic usage of sourcemap information (coverage tooling, stack trace sourcemapping, etc).

The LSP already works this way — it never formats before caching the source map. This flag lets build systems match that behavior.


whalenrp:
I started a discussion with a bit more context [over here](https://github.com/a-h/templ/discussions/1413) in case you'd rather discuss the right shape of the API first before merging. I realized formatting was just one part of the bigger problem and really it might make more sense to just do this as part of a build artifact anyway.

PR #1430, opened 2026-07-15
fix: add version and switch to using `pname`

vkcku:
The version was not added to the nix derivations which made it a bit harder to get the templ version from nix expressions directly. This fixes that by setting the version from the version file.

This also changes to setting `pname` instead of `name` directly. Using `pname` and `version` explicitly allows nix to generate the name of the derivation in the format: `${pname}-${version}`. This is the recommended approach as per the [documentation](https://nixos.org/manual/nixpkgs/stable/#sec-using-stdenv).

NOTE: I did not create an issue for this since it seemed like a small enough change to create a PR directly.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -45,7 +45,7 @@
           default = templ;
 
           templ = pkgs.buildGoModule {
-            name = "templ";
+            pname = "templ";
             version = builtins.readFile ./.version;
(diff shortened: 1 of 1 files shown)


a-h (approved):

a-h:
Thanks!

PR #1434, opened 2026-07-24
fix: dont delete dev-mode text files during a one-off generate

Napam:
## Problem

Running a one-off `templ generate` while a `templ generate --watch` dev server is up breaks every page the running dev server tries to render.

`--watch` writes each template's literals to a temporary `_templ.txt` file; a server run with `TEMPL_DEV_MODE=true` renders from those files. But `generate` called `deleteWatchModeTextFiles()` on every run, including non-watch ones which don't own these files. So a one-off generate deletes the `.txt` files of a separate running watch session, and every render then fails with a missing-file error. The watch session doesn't recover either. It skips rewriting files whose literals are unchanged.

**Repro:** start a watch-mode dev server (`templ generate --watch` plus an app run with `TEMPL_DEV_MODE=true` that renders from the `.txt` files), then run a plain `templ generate` (e.g. a CI/check step that verifies generated code is up to date). The one-off generate wipes the server's `.txt` files and every page then fails to render until the dev session is restarted.

## Fix

- `cmd.go`: only run `deleteWatchModeTextFiles()` in watch mode a one-off generate must not delete files it doesn't own.
- `eventhandler.go`: rewrite a `_templ.txt` file if it's missing from disk, even when literals are unchanged, so a watch session self-heals.

## Tests

Two regression tests in `main_test.go`; a non-watch generate leaves existing `.txt` files intact, and a deleted `.txt` file is recreated on the next regeneration.

## My use case

I run a dev server while also running static checks quite regularly (especially when using AI agents). Each check runs a plain `templ generate`, which was wiping the dev server's `_templ.txt` files and leaving it unable to render any page until I restarted it.

## AI disclaimer

I used Claude to help write the tests and suggest the fix.

### DIFF
--- a/cmd/templ/generatecmd/cmd.go
+++ b/cmd/templ/generatecmd/cmd.go
@@ -174,9 +174,16 @@ func (cmd Generate) Run(ctx context.Context) (err error) {
 		}
 	}
 
-	// Clean up temporary watch mode text files.
-	if err := cmd.deleteWatchModeTextFiles(); err != nil {
-		cmd.Log.Warn("Failed to delete watch mode text files", slog.Any("error", err))
+	// Clean up temporary watch mode text files. Only do this for watch
+	// sessions: a one-off generate must not delete the text files of a
+	// dev session that is still running, since its server renders from
+	// them (a missing file makes WriteString error out, failing every
+	// render, e.g. running `templ generate` while a `--watch` dev server
+	// is up).
+	if cmd.Args.Watch {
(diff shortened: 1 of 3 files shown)


joerdav:
I think this is an acceptable change to make. The reason this previously would happen is because the dev-mode files would be stored next to the templates themselves, so we needed a good way to clear them up before committing. Now they are located in temp directories this should be less of a requirement.

CC @a-h

